PR #3, opened 2016-12-05
Update README.md

jospoortvliet:
@schiessle @marinofaggiana pls review. If there is nothing bad, I suggest to let's get this live fast and add more with subsequent PR's.


schiessle:
great @jospoortvliet I just created almost the exact same README file 😄 I merged my small differences into your version.

I think that's a create initial README, @marinofaggiana feel free to extend it if you want anything to add.

marinofaggiana:
👍

Issue #29, opened 2016-12-30
Label of cancel option is way too small


rakekniven:
See red box on top right.
German word for "Cancel" is "Abbrechen". Given space for label is too small. Shortening function makes is unreadable for beginners. Same for other languages e.g. french AFAIK.

![2016-10-01 18-33-23 5689](https://cloud.githubusercontent.com/assets/2069590/21567242/fccf036a-ceaa-11e6-960c-3a1d899e53f1.png)

PR #35, opened 2017-01-14
Fix truncated text in cancel label in the login ViewController

rygen:
Fix for #29 . Removed the fixed width constraint (= 54) on the storyboard. The other changes were automatically done by Xcode (v8.2.1).


marinofaggiana:
Thanks @rygen

PR #36, opened 2017-01-16
Insert permission request for camera roll access in context

rygen:
A permission request for camera roll access is happening on app launch - together with notifications request - causing two alerts straight after each other, possibly leading to confusion. This commit removes it.
The permission is already being requested when access is really needed in CCMain.m (`PHPhotoLibrary requestAuthorization`), using the up-to-date iOS 8+ Photos framework, therefore making the call in AppDelegate.m unnecessary.

### DIFF
--- a/iOSClient/AppDelegate.m
+++ b/iOSClient/AppDelegate.m
@@ -188,19 +188,6 @@
     //[[AVAudioSession sharedInstance] setActive:YES error:nil];
     [[UIApplication sharedApplication] beginReceivingRemoteControlEvents];
     
-    // permission request camera roll
-    ALAssetsLibrary *lib = [[ALAssetsLibrary alloc] init];
-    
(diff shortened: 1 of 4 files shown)


marinofaggiana:
@rygen The request Authorization for camera roll it's necessary too for enable auto upload, please verify ... if i first do not openAssetsPickerController but  select Upload camera photos/videos ? 

thanks for verify

rygen:
True, it was not allowing the camera upload to be activated without going to the system settings first. Thanks for the heads up! 
The last commit fixes it. I've inserted a second permission request in context, so that it is requested when the "Activate camera upload" button is tapped. This request is also using the latest API (iOS 8+).

marinofaggiana:
mmm no, the permission must be check in Settings, view CCManageCameraUpload.m ;)

rygen:
Ahh, sorry. Didn't see there was another entry point to the settings for automatic upload. Hopefully now it's being requested in the right place and providing the user a good context for such request :)

marinofaggiana:
@rygen please wait, now i am worked for clear the code ;)

PR #59, opened 2017-02-06
Migrate to WKWebView

LukasReschke:
Signed-off-by: Lukas Reschke <lukas@statuscode.ch>


Kaitookidoo:
Hi i want to ask what about the cartiage files used are they also updated

Issue #47, opened 2017-01-27
[UX] Registration should allow bare hostnames, not require protocol


jlfranklin:
During account setup, the iOS client for Nextcloud asks for a URL.  If the user types just the hostname (e.g., cloud.example.org), then they are told "unsupported URL" and the field is cleared.

Multiple changes to the UX could be done here:

* Nextcloud/iOS should detect no protocol in the URL and try https, first, then http.  
* In the case that the hostname is wrong, leave the hostname in the field for the user to edit.  It's more likely they misspelled it or used the wrong TLD (.com instead of .org).
* The URL field could also be pre-filled with `https://` when the user selects the empty field.
* The error message should specify "The URL should start with http:// or https://"
* The field should use the URL keyboard.

marinofaggiana:
@jlfranklin Thanks for you help !

jancborchardt:
@jlfranklin exactly, it should be automatic! These are the tasks:

- [ ] Nextcloud/iOS should detect no protocol in the URL and try https, first, then http.
- [ ] In the case that the hostname is wrong, leave the hostname in the field for the user to edit. It's more likely they misspelled it or used the wrong TLD (.com instead of .org).
- [ ] The field should use the URL keyboard.

@nextcloud/ios this is an important first run UX fix.

tapanprakasht:
@marinofaggiana  I have fixed this issue. See pull request #96. But still the issue in open state

marinofaggiana:
yes !

https://github.com/nextcloud/ios/pull/96

jancborchardt:
@tapanprakasht awesome, thank you for your contribution! :) Looking forward to more from you.

jlfranklin:
The only thing left is setting the hostname field to use the URL keyboard.

rakekniven:
@jlfranklin Would you mind to file out separate issue for keyboard?

jlfranklin:
Sure. Added as #158.

PR #96, opened 2017-02-24
Registration should allow bare hostnames, not require protocol #47

tapanprakasht:
Implemented prefixing https protocol if no protocol is in baseUrl. If the connection is not successful url will remain in the textboxt for further editing.


marinofaggiana:
Very well @tapanprakasht !! 

Thanks a lot

PR #98, opened 2017-02-27
Remove old logic for ownCloud < 8

LukasReschke:
The old logic isn't required anymore and also seems to be buggy. If the version is "11.0.0" then `firstNumber` would be "1" which later in the comparison would fall into the else block and thus generate a link to the old invalid endpoint.

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>


oparoz:
:+1:

PR #99, opened 2017-02-27
Implement firebase

mario:
This implements basic Firebase for iOS. What this means is that we do not track anything specifically apart from Firebase does on it's own, and that's probably enough for the first version.

I have created a custom target called NextcloudCustom so Firebase will only be included when that target is being built. Later on, we can introduce a variable to disable firebase even if we're building the NextCloud custom.

Setting this up is pretty easy - replace GoogleService-Info.plist in the root of the project, and build NextcloudCustom.


marinofaggiana:
@mario it's possible do not use cocoapods but framework library ?

mario:
@marinofaggiana is there a technical reason for that?

Pods are easier to handle for the most part.

marinofaggiana:
I don't like cocoapods, better carthage or framework ... yes cocoapods are easier (for beginner) but add xcworkspace with several modify file project etc.

mario:
*shrug* We do want to lower the barrier to entry.

I will try to remove Pods tomorrow and just use Framework direct.

marinofaggiana:
Thanks @mario for users when the library static it's add with .h it's the same ...  

For beginner one only file : Nextcloud.xcodeproj is better  and simple  👍

mario:
Should be done now.

Issue #113, opened 2017-03-18
Missleasing translation to german


Dirk23:
There is one translation to German wich is misleading: The Switch in the automatic picture upload wich enables the background upload. It is translated to "Hintergrundbilder hochladen" wich is translated to english with: "Upload Backgroundpictures"!

I guess that is not what it's meant to be. It should be translated to German with:"Bilder im Hintergrund hochladen" wich means to upload Pictures in the background!

marinofaggiana:
@rakekniven ping

rakekniven:
Just changed it to "Bilder im Hintergrund hochladen" at transifex.
Will provide PR to improve english text.

PR #114, opened 2017-03-18
Fix translation according to issue #113

benediktg:
"Hintergrundbilder hochladen" → "Fotos/Videos von der Kamera im Hintergrund hochladen"
should fix issue #113


marinofaggiana:
@benediktg the issue of Translation please only in "Issue tab" and not in PR; all the database of Language is rewrite all days from Transifex service.

timbru31:
@benediktg you should be able to update the translation here, so the daily sync pushes the updated strings: https://www.transifex.com/nextcloud/public/

rakekniven:
Just changed it to "Bilder im Hintergrund hochladen" at transifex.

PR #176, opened 2017-04-13
Update Localizable.strings

rakekniven:
Fix reported on transifex
https://www.transifex.com/nextcloud/nextcloud/translate/#de_DE/ios-localizable/92030738?issue=yes


marinofaggiana:
thanks.

PR #302, opened 2017-07-14
Spelling: Shortened "_footer_crypto_cloud_"

comradekingu:
It is a major pain to translate strings of this length on the translation platform.
Edit: This isn't the whole string, so you end up having to scroll back and forth to read it. The only option is to export it to an external editor.
transifex
Splitting the string got voted down due to it having to do with encryption.


rakekniven:
I do not support to short this explanation. Limits of Transifex should not be the main decision to shorten strings. The explanation regarding encryption is important in my opinion. 
Translated string to german myself and it works. For sure Transifix UI could be improved to handle bigger strings.

comradekingu:
@rakekniven The main reason for shortening the string is to cut down on fluff, which considering this is well over twice as short, there is quite an amount of. It is hard not to.
What part of explaining encryption is missing?

marinofaggiana:
Hi boys, other opinion ? 😄

rakekniven:
@comradekingu Sometimes it is more than just technical explanation. E.g. the words about encrypting folder names and other words are just fine. It is just my opinion.

comradekingu:
@rakekniven The main reason to shorten it is because it is abhorrently redundant, and badly written at that. What part of it is missing, and if so, could you cite it? I made an effort to make sure all the info was still there.

Issue #293, opened 2017-07-02
Can't Preview/Share Downloaded Mobi Files. iOS 11?


Jon-Schneider:
Device: iPad Air 2 running iOS 11 Beta 2
If I download a mobi file within the app by clicking on the file in the file nav, it will display the download progress but it will not open a "Preview" (which in OwnCloud is the placeholder icon preview for an unknown file type) after the download completes, or if I click on the file again after downloading. If I swipe on the filename and select the "Open In" option from "More" I have the same problem - the download will complete but the preview will not open, and the system share menu never opens.

Edit: 9/12/17
Further testing revealed that this appears to affect both iPads (but not iPhones) running iOS 10 and iOS 11, and it affects sharing of all files from the Main File menu, not just files without previews - it's just more obvious with files that can't be previewed because it should open the Share menu automatically.

marinofaggiana:
Thanks ! Can you test last beta ?

marinofaggiana:
iOS 11 is in beta version.

Jon-Schneider:
Still present in iOS 11 GM. Fixing in PR.

PR #371, opened 2017-09-13
Fixes System Share menu not consistently presenting from CCMain

Jon-Schneider:
Background:
I first noticed this when I reported Issue #293. Further testing revealed that this affects sharing of all files from the Main File menu, not just files without previews - it's just more obvious with files that can't be previewed because it should open the Share menu automatically.

This affects both iOS 10 and 11.

Fix:
When showing System Options Menu from CCFavorites and CCMain, now passes the superview as the view the menu should be shown in. It looks like there's a system issue with trying to present the options menu from a view's frame within that view.

### DIFF
--- a/iOSClient/Favorites/CCFavorites.m
+++ b/iOSClient/Favorites/CCFavorites.m
@@ -392,7 +392,7 @@
         _docController = [UIDocumentInteractionController interactionControllerWithURL:url];
         _docController.delegate = self;
         
-        [_docController presentOptionsMenuFromRect:self.view.frame inView:self.view animated:YES];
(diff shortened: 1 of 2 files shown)


Jon-Schneider:
@marinofaggiana Do you know when you’ll have a chance to review?

marinofaggiana:
Sorry @Jon-Schneider I am very busy for the new version ( look levelop branch ) then I apply your fix !   Thanks for your precious help, try the develop version ;-) but with xcode9

rakekniven:
@Jon-Schneider Can you resolve conflicts of your branch? Thx.

Jon-Schneider:
This is no longer an issue for me.

PR #388, opened 2017-10-03
Fix modified time

### LINKED ISSUE #256: auto upload created and modified date get set to upload date/time
while manual upload seems to preserve created and modified

for me at least it would feel natural to have auto upload preserve dates as well

cheers

mario:
Should fix #256, at least the part about preserving mtime. Didn't look into filename creation.

Signed-off-by: Mario Danic <mario@lovelyhq.com>

### DIFF
--- a/iOSClient/Networking/CCNetworking.m
+++ b/iOSClient/Networking/CCNetworking.m
@@ -1262,6 +1262,7 @@
     NSString *fileNameForUpload;
     
     if ([CCUtility isFileNotCryptated:fileName]) fileNameForUpload = sessionID;
+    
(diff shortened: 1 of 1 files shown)


marinofaggiana:
Required modify but ok try a test with add : 

[request setValue:[NSString stringWithFormat:@"%ld", since] forHTTPHeaderField:@"X-OC-Mtime"]

I don't know X-OC-Mtime, where is the specification ?

mario:
@marinofaggiana unfortunately nowhere, but this works - I know since I fixed the same thing for Android ;)

"since" is basically time in miliseconds since 1970.

PR #391, opened 2017-10-07
Issue 340 file size for small files shows decimal point

### LINKED ISSUE #340: file size for small files shows decimal point
if i create a small text file which is, for example, only 67 bytes in size, the size shows as 67.00 bytes.

i'm not sure how i could have a fraction of a byte :-) 
the decimal portion should be removed.

cheers, wizdude.

matthiasdanner:
This fixes #340 and allows for file sizes greater than 1024 TB ;)

### DIFF
--- a/iOSClient/Utility/CCUtility.m
+++ b/iOSClient/Utility/CCUtility.m
@@ -473,7 +473,7 @@
 + (NSString *)transformedSize:(double)value
 {
     int multiplicationFactor = 0;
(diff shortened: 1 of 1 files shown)


gfoldv:
Should check against the fraction part of the division instead of the multiplicationFactor as index. Then it would also work for whole number MB, GB, etc.

matthiasdanner:
Now using numberformatter for dynamic length of the fractional part

Issue #359, opened 2017-08-23
use a better default naming scheme for instant upload


mbiebl:
The current default naming scheme seems to be
YY-MM-DD HH-mm-ss counter.jpg

I'd like to suggest a few changes:
a/ Use a full year, i.e YYYY
b/ Avoid using spaces
c/ Don't use the same format for day and time.
  Example: 17-06-01 16-04-02 0001.jpg is very hard to decypher, as you don't know which part is which.

Suggested change: YYYY-MM-DD_HHmmss_0001.jpg
Example: 2017-04-02_160402_0001.jpg

mbiebl:
The focus is on the date, which imho is the most important information.

rakekniven:
Dup of #332 
I prefer to close older issue as description here is better IMHO.

rakekniven:
Related to #325 as well.

doomi:
Is there a reason to not use the iOS filename?
Almost every other auto image upload software has an option to either use original name (IMG_xxxx.jpg) or rename to timestamp.

doomi:
Still not fixed. IMO it should use the iOS filename by default. But if not, then use a 4 digit year as default and certainly no spaces in the filename.

dennore:
Please remove option Subfolders in autoupload for YYYY/MM/photo.jpg

and let us add custom filenames with path because that brings all the options we need.

e.g. i would like to have all photos from one year in one folder:
YYYY/YYYY-MM-DD hh-mm-ss [originalfilenamecounter].jpg
which gives me an error: / is an invalid character 👎

rakekniven:
@dennore That would be a different issue IMHO.

laurentperrinet:
The suggested change: YYYY-MM-DD_HHmmss_0001.jpg  would be great and would follow the standards from

https://en.wikipedia.org/wiki/ISO_8601

mbiebl:
> Still not fixed. IMO it should use the iOS filename by default. But if not, then use a 4 digit year as default and certainly no spaces in the filename.

I think this is sufficiently addressed already in the App where you have the option to use the original file name: 

<img width="886" height="1908" alt="Image" src="https://github.com/user-attachments/assets/a65d8e86-bb01-4792-b2c5-a2a0cbe8f49b" />

(sorry for the German locale)

"Ursprünglichen Dateinamen behalten" means to keep the original file name.

PR #407, opened 2017-11-07
Improve auto upload file mask

c4539:
This change makes the separator before and after the time string configurable.

### DIFF
--- a/iOSClient/Create/CCCreateCloud.swift
+++ b/iOSClient/Create/CCCreateCloud.swift
@@ -452,7 +452,7 @@ class CreateFormUploadAssets: XLFormViewController, CCMoveDelegate {
         
         if let valueRename = valueRename {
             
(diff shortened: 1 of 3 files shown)


c4539:
This should close #359

c4539:
@marinofaggiana Do you think this can be included at some point? With one of the last updates, the nomenclature of the Auto-Upload files has been changed and since then, it's no longer possible to have a space after the timestring (which used to be the default). Thanks.

marinofaggiana:
@c4539 Simon change te code only for improvement but not for modify the actual "default" filename ... the users do not want changes

c4539:
@marinofaggiana how about changing the trim only? Shouldn’t change anything to the default configuration but enables users to use a white space as the separate between the timestring and the file number.

marinofaggiana:
@c4539 try ;)

c4539:
@marinofaggiana Thanks a lot, works like a charm ;-)

PR #466, opened 2018-01-14
Force remove folder lock

comradekingu:
Not sure about this one, I think it is an action (?)
The alternative would be: "Forced removal of folder lock"


comradekingu:
@marinofaggiana Could you explain what happens?
"Forced remove" is not correct.

marinofaggiana:
If the e2e folder remain locked (issue) you can remove (forced) the lock ...

PR #613, opened 2018-07-09
Show toolbar for external sites

MorrisJobke:
@marinofaggiana I got this as patch to solve the issue, when you have an external site configured and you want to close the web view there was no close or back button. With this change there is one.

Does that make sense?


marinofaggiana:
yes, but not here ... https://github.com/nextcloud/ios/commit/478dcf4c5cb9ffc14b9bcd29c65e5ae21e265e9c

PR #694, opened 2018-10-16
[DE] Grammar Improved

Knot3n:
German sentences slightly grammatical improved, adapted to the pronunciation. 
Punctuation added, sentences truncated.


jancborchardt:
Hey @Knot3n, as said in the other issue we do translations in Transifex. You need to get an account and then open https://www.transifex.com/nextcloud/nextcloud/ios/

The format is not the same (as far as I know) so you have to copy over the text, as it will need to be reviewed too. Maybe there is an import function but I don’t know.

(FYI I just checked the readme and there it also says we use Transifex: https://github.com/nextcloud/ios#how-to-contribute – sorry for the confusion. :\)

Issue #725, opened 2018-11-16
After selection of more than 6 files iPhone photos error


KurtCa:
### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Nextcloud app opening
### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
The application terminates 
### Steps to reproduce                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Select more than 6 photos in the iphone photos app.
then share with nextcloud app.
The nextcloud app opens and terminates instantly.
If i select less than 6 photos, the app opens as expected.
### iOS version
IOS12
### App version
2.22.5
### Server configuration

**Operating system**:
ubuntu  18.04.1 LTS
**Web server:**
apache
**Database:**
mysql 5.7.24
**PHP version:**
7.2.12
**Nextcloud version:** (see Nextcloud admin page)
14.0.3

tjmc:
I'm able to reproduce this on my iPhone SE (12.1). 

6 photos is the magic number to push the share extension over the memory limit.

I'm seeing:

` thread #1, queue = 'com.apple.main-thread', stop reason = EXC_RESOURCE RESOURCE_TYPE_MEMORY (limit=120 MB, unused=0x0) `

when the share extension is killed.

My thoughts would be to resize the UIImage inside ShareViewController using NCUtility to avoid reaching the memory limit. I'm not sure what width should be though. Maybe a hardcoded value (500)?

`image = [NCUtility.sharedInstance resizeImageWithImage:image newWidth:500];`

marinofaggiana:
Thanks @tjmc try with your modify, I don't remember why 500 ... I think that is a old misprint o test ..

PR #755, opened 2018-12-18
Use resized images for share extension to reduce memory usage

tjmc:
6 photos is the magic number to push the share extension over the memory limit. Resizing the images for use to reduce memory usage.

Fix for #725 - After selection of more than 6 files iPhone photos error


marinofaggiana:
Good @tjmc thanks for your job !!

Issue #636, opened 2018-08-21
Change timestamp of older folders 


TecJon:
#### Timestamp
![date](https://user-images.githubusercontent.com/11375752/44357977-e3a2d200-a4b3-11e8-99d8-0321ca94e6a6.jpg)

The text below folder "xx days ago" is useful for recently changed/created folders. However, as soon as every folder was modified/created longer than 30 days ago, the info is not useful anymore. If the date is longer than 30 days ago it should tell the date itself: "modified 07/08/2015" (in the date format specified in the iOS settings).

rakekniven:
My idea: Do it like the web UI is doing it:

![2018-08-27_10h57_08](https://user-images.githubusercontent.com/2069590/44650827-fe130900-a9e7-11e8-84da-83e60595a80f.png)

rakekniven:
@jancborchardt What is your opinion here?

jancborchardt:
For files older than a year, it should show the date in the date format specified in iOS settings, but human readable (That is, like "September 7 2018" or so.

For files edited more recently than one year ago, it should use the relative dates as in the web interface.

wolframroesler:
IMHO even "8 months ago" is too fuzzy. I suggest to show the age ("13 days ago" etc.) for files younger than a month, as before, and replace the "over 30 days" message by the (localized, human-readable) date.

Working on it, wait a sec, pull request pending.

marinofaggiana:
@wolframroesler available in develop branch, build 14 👍

PR #756, opened 2018-12-21
Show modification date for files older than 30 days

wolframroesler:
As mentioned in #636, for files older than 30 days the iOS client now displays the file modification date rather than the "over 30 days" message (in human-readable, locale-specific format).

bildschirmfoto 2018-12-21 um 10 14 24

### DIFF
--- a/iOSClient/Main/NCMainCommon.swift
+++ b/iOSClient/Main/NCMainCommon.swift
@@ -283,7 +283,7 @@ class NCMainCommon: NSObject {
             } else {
                 
                 cell.imageItem.image = image
-                cell.labelInfo.text = CCUtility.dateDiff(metadata.date as Date) + " " + CCUtility.transformedSize(metadata.size)
+                cell.labelInfo.text = CCUtility.dateDiff(metadata.date as Date) + ", " + CCUtility.transformedSize(metadata.size)
                 
                 // image Local
                 let tableLocalFile = NCManageDatabase.sharedInstance.getTableLocalFile(predicate: NSPredicate(format: "fileID == %@", metadata.fileID))
@@ -537,7 +537,7 @@ class NCMainCommon: NSObject {
                 let iconFileExists = FileManager.default.fileExists(atPath: CCUtility.getDirectoryProviderStorageIconFileID(metadata.fileID, fileNameView: metadata.fileNameView))
                 
                 // Lable Info
-                cell.labelInfoFile.text = CCUtility.dateDiff(metadata.date as Date) + " " + CCUtility.transformedSize(metadata.size)
+                cell.labelInfoFile.text = CCUtility.dateDiff(metadata.date as Date) + ", " + CCUtility.transformedSize(metadata.size)
                 
                 // File Image
                 if iconFileExists {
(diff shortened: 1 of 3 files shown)


jancborchardt:
@wolframroesler very nice, thanks for the initiative of directly submitting a pull request! Welcome to Nextcloud! :tada: 

Only one tiny piece of feedback: Currently the comma placement is a bit off

> Aug 3, 2016 165 bytes

mixes the year and the file size. Instead it could be:

> Aug 3 2016, 165 bytes



Also @nextcloud/android what do you think about this? We should do the same across our mobile apps.

jancborchardt:
Also @TecJon @rakekniven please review as this fixes the issue https://github.com/nextcloud/ios/issues/636 you were involved in too. :) And cc @nextcloud/ios for review.

tobiasKaminsky:
On Android we rely on android's built in relative timestamp:
- show x days ago (until 7 days)
- within the same year, show only "03. March"
- otherwise show complete date, e.g. 20.12.2013

wolframroesler:
@jancborchardt thanks for your reply.

> Only one tiny piece of feedback: Currently the comma placement is a bit off

You're right, there should be a comma between the date and the size, but then there should also be a comma between the age and the size (for files younger than a month) to make it consistent. Will look into it.

As for the comma between date and year, that one comes from the locale-specific date formatting function (`setDateStyle:NSDateFormatterMediumStyle`) so we've got little control over that, and little chance to get rid of it (unless we do some ugly string manipulation stuff which may or may not work in other locales). Here's how it looks like when the phone is switched to German:

bildschirmfoto 2018-12-21 um 16 35 45

wolframroesler:
@tobiasKaminsky 

> * show x days ago (until 7 days)
> * within the same year, show only "03. March"
> * otherwise show complete date, e.g. 20.12.2013

Does anyone know if iOS has a similar functionality? It would be nice if it were consistent with the Android version, however we need to use some operating system functionality in order to get locale-specific formatting right.

wolframroesler:
@jancborchardt by the way, thanks for the welcome, allow me to introduce myself. I'm a veteran software developer, located in Germany, doing C++ on Linux for a living, self-hosted Nextcloud in my bedroom, and the "over 30 days" thing has been bugging me for a long time. Decided to fix it in my winter vacation, was happy to find that someone else had already filed an issue for it. Nextcloud is a central part of my private IT infrastructure, I'm glad to contribute and thus return a bit to the community.

TecJon:
The screenshot with the comma looks good to me.

jancborchardt:
Nice @wolframroesler! @marinofaggiana is the @nextcloud/ios lead and we could really use more community contributors on the iOS app, so any contribution is very very welcome! :)

Also cc @Ivansss @nextcloud/ios for review

marinofaggiana:
@wolframroesler 👏 good job ! You are welcome and your help is much appreciated !

wolframroesler:
@marinofaggiana Thanks! My resources are limited by I'm glad to be of service.

Are there already plans for when the next version of the iOS app will be released?

marinofaggiana:
@wolframroesler January 2019 in TestFlight

Issue #734, opened 2018-11-26
Wrong Translation in German iOS App


quotengrote:
### Expected behaviour      
"WLAN-Netzwerk wird benötigt um "NUMBERofITEMS" Elemente hochzuladen"                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   

### Actual behaviour  
"WLAN-Netzwerk benötigt  "NUMBERofITEMS" Elemente zum Hochladen"                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   

### Steps to reproduce                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
In Nextcloud App for iOS in the menu: "Mehr" there is menu called "Übertragungen".
If i have upload only when WLAN is active activated and there are items to upload there is shown a text: 

NUMBERofITEMS is actually a Number and not the string itself ;-)

### iOS version
12.1
### App version
2.22.5.10

wolframroesler:
Current version (2.22.7) has (line 356 in iOSClient/Supporting\ Files//de.lproj/Localizable.strings):

> "_tite_footer_upload_wwan_"     = " WLAN-Netzwerk benötigt %lu %@ zum Hochladen";

Seems OK to me (%lu is a number). Can you check if the problem still exists, or upload a screenshot that shows the exact problem?

BTW, your suggested solution is missing a comma, it should be "benötigt um ... hochzuladen".

wolframroesler:
Working on it. Will also fix a couple of other inconsistencies in the German translation (for example, there's a second message, identical to the one you mention but about download, which says "Wi-Fi" instead of "WLAN").

wolframroesler:
Ok, changing the translation in the source was wrong, they need to be edited on https://www.transifex.com/nextcloud/nextcloud/ios/. Did it, hopefully they'll show up in the next version.

PR #757, opened 2018-12-21
Several fixes in the German translation

wolframroesler:
Fixed the "need Wi-Fi to upload ... elements" message which was semantically wrong (issue #734). Also added a couple of missing commas and made a few texts more consistent (e. g., don't call it WLAN here and Wi-Fi there). Added missing space before ellipses ("Loading ..." instead of "Loading...", the latter has a different meaning).


jancborchardt:
@wolframroesler all our translations are done via Transifex: https://www.transifex.com/nextcloud/nextcloud/ios/ – only the English original is something we change here, but the translations are periodically loaded from there. Could you make the changes there instead?

And sorry for the confusion!

wolframroesler:
@jancborchardt Replicated my changes on Transifex. Can you check if it worked?

jancborchardt:
@wolframroesler thanks a lot! I’m not really active on our Transifex – @MariusBluem since you use the iOS app (maybe in German), can you check? :)

Issue #342, opened 2017-08-09
share links contain index.php


wizdude:
when i share a file and select "send link to..." and select E-mail, the resultant link in the e-mail contains a reference to index.php. i have configured my nextcloud not to use index.php.

example when i share picture from ios client:
https://cloud.blah.com.au/index.php/s/GJ4Hhw4RVhdFfGx

example when i share the same file from web client:
https://cloud.blah.com.au/s/GJ4Hhw4RVhdFfGx

i would prefer not to have the reference to index.php.

cheers, wizdude

gfoldv:
If the server API does not provide this information - I guess not - then it would be nice to have an entry in the settings. Like a switch "Pretty URLs".

Ricain:
Hello, no news for this issue?

Thank

PR #874, opened 2019-06-01
Using URL's returned from the server for sharing

JamieWhite:
Doing this so the app makes use of pretty URL's without `index.php`

At the moment the URL generation is hard coded on the client. Seems like a more sensible idea to use the server returned `url`. I left the hard coded version in as a fallback as Im not sure if it's still needed.

Fixes this issue https://github.com/nextcloud/ios/issues/342


marinofaggiana:
@JamieWhite Great !!!! Thanks for your job, soon merge in develop for the next version !

PR #892, opened 2019-06-28
Provide a banal 'paste' postmessage implementation.

mmeeks:
Browsers will allow apps to populate the clipboard, but not read
from it without user input. This patch makes it possible for
context menus with 'Paste' in them to trigger a real OS paste
when hosted inside the webview.


marinofaggiana:
@mmeeks, thanks a lot !!

PR #894, opened 2019-07-01
"Delete account" -> "Remove account"

leonklingele-work:
"Delete account" sounds more like you would want to delete your whole Nextcloud account on the server whereas "Remove account" sounds like you intend to remove it locally only.


marinofaggiana:
Thanks @leonklingele-work. done

PR #928, opened 2019-08-28
Improve scroll performance

JamieWhite:
The scrolling performance is pretty choppy at the moment. Looking through instruments I can see there's a few bottle necks where we're tinting images so I've cached the images after they've been tinted.

I'd quite like to have another look at this to get the scrolling more performant but these were some quick wins.

Screenshot 2019-08-28 at 21 27 03


marinofaggiana:
@JamieWhite oh! great job !! now I look it 👍

marinofaggiana:
Available 2.24.0 Build 13. Thanks a lot @JamieWhite

PR #933, opened 2019-09-05
Creating image cache

JamieWhite:
I noticed we're loading files from disc on the main thread and not caching them.

I've created a class that relies on `NSCache` for caching so it'll automatically purge images as system resources get low. I've assumed `metadata.fileID` is unique so i've used that as the key for the cache.

Additionally I'm loading these images asynchronously off the main thread which helps with scrolling performance.

If you test this in a folder full of images it still stutters a bit as it's loading but once they've all been loaded into memory it's really quick.

Next task is to find out why it's still stuttering as the images are loaded into memory 🤔

### DIFF
--- a/Nextcloud.xcodeproj/project.pbxproj
+++ b/Nextcloud.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


marinofaggiana:
Hi @JamieWhite great improvement !! do you have make some test for know the real convenience ?

marinofaggiana:
Path applicate

Issue #1044, opened 2019-12-17
Remove manually imported lib and manage them with carthage


immortal79:
### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Frameworks should be imported with carthage.

### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Some frameworks are in the 'Library' folder for example 'XLForms' (V 3.3.0) could be imported and managed with carthage. 
Is there a particular reason doing this way ? Are some of those libs modified ?

### iOS version
Any
### App version
Any

PR #1051, opened 2019-12-20
XLForm as Carthage dependency

immortal79:
Can help for future dev and bug fix

Fix for #1044 


marinofaggiana:
I use an XLForm modified, but this year I want remove the differences, for now I cant use the standard lib.

immortal79:
I didn't see problems while using the app with it as dependancy. In what way is it modified ?

marinofaggiana:
the fonts of picker in CCManageAccount but the view will be rewrite with a new UI in according with Nextcloud Talk.

PR #1053, opened 2019-12-20
Check dark mode at first launch - Fix #1052

### LINKED ISSUE #1052: Dark mode is not enabled according to system preferences at first launch
### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
After logging dark mode should enabled if enabled system wide.
### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
At first launch, the app doesn't check if dark mode should be enabled which results in the app being white instead of dark even if dark mode is enabled in the system settings.

### Steps to reproduce                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
(enable dark mode)
Start with a newly installed app / Uninstall the app.
Log-in
The app is white instead of dark.

### iOS version
13.3
### App version
Latest

immortal79:
Ensure that dark mode is enabled according to system preferences.
Fix for #1052

### DIFF
--- a/iOSClient/Utility/CCUtility.m
+++ b/iOSClient/Utility/CCUtility.m
@@ -659,6 +659,11 @@
 
 + (BOOL)getDarkMode
 {
+    if ([self getDarkModeDetect]) {
+        if (@available(iOS 12.0, *)) {
+            return [UIScreen mainScreen].traitCollection.userInterfaceStyle == UIUserInterfaceStyleDark;
+        }
+    }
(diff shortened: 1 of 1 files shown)


marinofaggiana:
Hi @blackivory86, simply, the Nextcloud iOS is available for >= iOS 10 that do not support the  userInterfaceStyle, so the user can select the Dark Mode (color) manually and set this option in the keychain, so under iOS 13 I use the same logic.

Issue #1057, opened 2019-12-28
iOS 13 simulator: Access to UITextField's _placeholderLabel ivar is prohibited


blackivory86:
### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
I can launch the app in the simulator and login to test the app.

### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
When opening the login form the app crashes with the follwing log message:
```
*** Terminating app due to uncaught exception 'NSGenericException', reason: 'Access to UITextField's _placeholderLabel ivar is prohibited. This is an application bug'
```

### Steps to reproduce                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             

Launch the App in simulator with iOS 13.3 SDK, Tap the white login button

### iOS version
iOS 13.3 simulator

### App version
*develop:* commit 3901328
*master:* commit aeb4cc5

### Server configuration

**Operating system:**

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)

blackivory86:
If you want I could create a PR to fix this.

blackivory86:
Nevermind. Was totally thrown off by the github UI not giving any indication that I had already forked the repo exactly a year earlier at 35c3 and working with code from a year ago

PR #1058, opened 2019-12-28
#1057: Replace private Uitextfiel.placeholder API from Login screen

blackivory86:
Fix for #1057

Removes access to '_placeholderLabel' which seems to be "private API" with attributed strings to set the placeholder in the desired color.

There are more usages of `_placeholderLabel` in `XLFormTextView` which is a third party dependency, so I did not touch it with this PR.


blackivory86:
I'm a litte confused. The README says PRs should be against `master`, but LoginView seemed more up2date on develop so I decided to do the PR against that. Can you clarify what'
s the correct approach?

blackivory86:
Nevermind, this is embarassing  
I did not realize I had already forked this repo a year ago at the same time and was exactly a year behind. Of course it's already 2019 and not 2018 🤦‍♂ 

This PR is probably completely obsolete. Will check and then probably close this.

PR #1067, opened 2020-01-03
nextcloud ios app album layout, sort by folder

xuansongzhang:
the album in ios app called media, sort the photos by date, while nextcloud web album sort the photos by folder. 
could you please add by folder layout, and let album can be switch between by folder and by date.


marinofaggiana:
please create an issue for request.

PR #1080, opened 2020-01-08
Fix _delete_account_ for french

JorisBodin:
Hello,

In french this translation is better

And there's a lot of unused keys. If we clean all the keys, would you be okay?
image

Best regards :)


marinofaggiana:
Hi @JorisBodin you can't change here the translation but only here: 

https://www.transifex.com/nextcloud/nextcloud/

Thanks a lot for your help

JorisBodin:
Ok :)
But for remove unused key, it's only from the git?

PR #1101, opened 2020-01-24
Menu displayed as drawer

immortal79:
I replaced the actual menu everywhere with a more intuitive, accessible and modern bottom drawer.
New icons were added to better represent possible actions.

When creating a new document :
Simulator Screen Shot - iPhone 11 - 2020-01-24 at 14 46 42

When clicking (...) for a document :
Simulator Screen Shot - iPhone 11 - 2020-01-24 at 14 46 37

Select menu :
Simulator Screen Shot - iPhone 11 - 2020-01-24 at 14 46 48

PR #1112, opened 2020-01-27
Newdrawer fix for #1101

immortal79:
- To prevent any weird behaviour the menu is now closed on rotation. 
- This branch also provides fix for devices < iOS 11.
- ReMenu lib has been removed from sources.


marinofaggiana:
he he he : @immortal79 

To prevent any weird behaviour the menu is now closed on rotation, but it is too easy

immortal79:
I know ... x) but I think it's good enough as the probability that the user would rotate his device with the menu open is very low.

I had an other solution but the frame resizing wouldn't follow the rotation: the frame was resized after rotation which also didn't look that good. 

The ground problem is that the new menu is a ViewController and when presented it takes the whole screen on iPad and Max/Plus sizes (it covers both the Master and Details ViewController) so it has to be resized manually. 

I know I was the one who introduced this new lib but let me know if you have any idea on how I could fix this ?

PR #1127, opened 2020-01-30
correction user agent for OnlyOffice integration

LinneyS:
https://github.com/ONLYOFFICE/onlyoffice-nextcloud/issues/179#issuecomment-580238212


marinofaggiana:
OK, I have inserted manually this, will be available for the next build (3) in TestFlight

PR #1205, opened 2020-04-08
replace $amp; with real & symbol in intro screens

valentindey:
replace $amp; with real & symbol in intro screens

Signed-off-by: Valentin Deyringer <v.deyringer@tuta.io>


marinofaggiana:
Hi @valentindey for the translation we use transifex:

https://www.transifex.com/nextcloud/nextcloud/dashboard/

you can use this link for the modify. 

Thanks
m.

PR #1208, opened 2020-04-12
removed deprecated viewDidUnload implementations

jamesstout:
From docs: Deprecated - Views are no longer purged under low-memory conditions and so this method is never called.

Signed-off-by: James Stout <stoutyhk@gmail.com>

### DIFF
--- a/iOSClient/Library/MWPhotoBrowser/MWPhotoBrowser.m
+++ b/iOSClient/Library/MWPhotoBrowser/MWPhotoBrowser.m
@@ -285,19 +285,6 @@ static void * MWVideoPlayerObservation = &MWVideoPlayerObservation;
     
 }
 
-// Release any retained subviews of the main view.
-- (void)viewDidUnload {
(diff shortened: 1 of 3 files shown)


marinofaggiana:
Hi @jamesstout this view is abandoned look this new V 3 in progress:

https://github.com/nextcloud/ios/tree/Coherence

Thanks a lot
m.

tarasis:
@marinofaggiana may I suggest putting the information about v3 Coherence in the project readme.md. I assume then that main is just the releases, develop is now intended for just fixing critical bugs and triage, and anything else should be on Coherence? 

I was about to do a similar thing as @jamesstout, but with the depreciated use of UIAlertView in CCDetail.m. Figuring I should test behavior first it highlighted a bug with password protected PDF's (if you just enter a 0 length password you keep getting the password dialog back (and wrongly resigns the keyboard), its not obvious you have to enter something that fails for the dialog to stop re-appearing)

PR #1209, opened 2020-04-12
Use static NSDateFormatter where possible

jamesstout:
Per Apple: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DataFormatting/Articles/dfDateFormatting10_4.html#//apple_ref/doc/uid/TP40002369-SW10

> Creating a date formatter is not a cheap operation. If you are likely to use a formatter frequently, it is typically more efficient to cache a single instance than to create and dispose of multiple instances. One approach is to use a static variable.

I've added performance and equality tests. All equality tests pass and performance in these simple tests show 1200% speed up and 2100% fewer heap allocs.

The new tests for CreatingNewAssetsToUpload using a device with ~28000 assets, show a 14% improvement (8 mins vs 9 mins) with the static formatter.

Signed-off-by: James Stout <stoutyhk@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ UserInterfaceState.xcuserstate
 ## Carthage
 Carthage/
 Cartfile.resolved
+/Nextcloud.xcodeproj/xcshareddata/xcbaselines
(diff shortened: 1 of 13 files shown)


marinofaggiana:
Hi @jamesstout I'll have a look soon, thanks a lot !!!

PR #1252, opened 2020-05-12
update afnetworking framework to the latest version

zhanleewo:
update afnetworking framework to the latest version


marinofaggiana:
Hi, thanks but we work with a new lib without AFNetworking, soon will be removed.

Issue #1218, opened 2020-04-17
Design refresh


JorisBodin:
Hello,

After discussion with @marinofaggiana and @jancborchardt, here are our mokups to refresh the current design.
The iOS and Android apps will be aligned designwise.
https://infomaniak.invisionapp.com/share/AQWL9RLZR6G#/410807685_App_Nextcloud_IOS_-_Home

In the new design only the header bar is going away, the rest should still be themed.

Android: https://github.com/nextcloud/android/issues/5207

jancborchardt:
Perfect, thanks a lot @JorisBodin! :)

jancborchardt:
This was mostly done back then, so let’s close it. :)

PR #1254, opened 2020-05-12
New tab bar

immortal79:
This PR contains the new tabbar discussed in #1218 .
New icons will come in a later PR.
Simulator Screen Shot - iPhone X - iOS 13 - 2020-05-12 at 15 33 04


marinofaggiana:
good @immortal79 but 2,3 time of 10 touch in the (+) take the focus the cell on tableview in background ... it would be possible to increase the focus zone of the button + ?

Issue #1246, opened 2020-05-11
New navigation bar


marinofaggiana:
@immortal79 good job !!

look the firs issue: 

https://customer.swing.cloud/index.php/s/CafzSsJ486Cp9kE

immortal79:
I will work on those bugs. 
If you find others, update the issue and will try to also fix them.

immortal79:
https://github.com/nextcloud/ios/blob/7fbaa9ff0a6870acb9563092e7ea423c217bc437/iOSClient/AppDelegate.m#L1019

I don't understand this part of the code and I think it's the one creating the bug. Why is it removing the view and then adding it back again ?

marinofaggiana:
Yes was for the quick look navigation bar color but since it creates problems I have removed it in this moment

marinofaggiana:
Note: I have added Notification in "More" menu but soon I think that replace the "Favorites" in tabbar

marinofaggiana:
@immortal79 

with information folder the search field on up scroll and disappears 

![20-05-11 16-50-03 0039](https://user-images.githubusercontent.com/8616947/81576115-23a49980-93a8-11ea-902f-12a26e95f982.png)

but not without  information folder

![20-05-11 16-50-18 0040](https://user-images.githubusercontent.com/8616947/81576269-58b0ec00-93a8-11ea-96a2-8862aa9ac8fd.png)

marinofaggiana:
Sorry, do not depends from the "Information folder" but from the first load and the re-view

marinofaggiana:
Fix for #1246, it seems that this is a bug with iOS 13 and our tableview layout.

@immortal79  remains this bug

marinofaggiana:
@immortal79 remains the bad refreshControl

marinofaggiana:
![Schermata 2020-05-26 alle 15 44 01](https://user-images.githubusercontent.com/8616947/82908121-d1e33e00-9f67-11ea-9738-5663dcf188fb.png)

@immortal79 several time the now the "Sorted by name is cut"

marinofaggiana:
https://github.com/nextcloud/ios/pull/1273

PR #1262, opened 2020-05-18
Fix refresh control

immortal79:
It looks like it could break the intended refresh behaviour but it doesn't change anything in my tests. 
Maybe it broke something somewhere else ?
Ref: #1246


marinofaggiana:
I make some test.

PR #1263, opened 2020-05-18
New icons for tab bar

immortal79:
Here are some new icons to go along with the new tab bar, I don't know if they are gonna suit you.

Here is what it looks like:
Simulator Screen Shot - iPhone X - iOS 13 - 2020-05-18 at 11 44 35


marinofaggiana:
This is material for @jancborchardt :-)

immortal79:
I added more new icons. They come from the default iOS library and they make the app overhaul more coherent.

Simulator Screen Shot - iPhone X - iOS 13 - 2020-05-19 at 15 14 46
Simulator Screen Shot - iPhone X - iOS 13 - 2020-05-19 at 15 14 49

marinofaggiana:
Note: the icons used in other views as for example the activity must be a vector pdf with large dim.

marinofaggiana:
Simulator Screen Shot - iPhone 11 - 2020-05-20 at 10 03 39

@immortal79 @jancborchardt and remove the title ?

jancborchardt:
> @immortal79 @jancborchardt and remove the title ?

No, these menu icons need text to make it obvious what it is. Otherwise we have a case of https://en.wikipedia.org/wiki/Mystery_meat_navigation

immortal79:
> Sure thing @immortal79, looks super nice and iOS-native! :)
> 
> Only one thing: The "Share" icon looks more like an "Open with" icon – does iOS have this triangular "Share" icon as well?
> 

I agree with you this icon can be misleading but the actual/triangular one is a bit bold to fit the theme, I will try to find a better fitted one.

immortal79:
> Note: the icons used in other views as for example the activity must be a vector pdf with large dim. But is not a problem because I check that before the merge

I can export with larger dim but why do that ? As PDF is vector based this shouldn't make a difference ?

jancborchardt:
> I agree with you this icon can be misleading but the actual/triangular but it's actually used in the entire system (maybe a bit too much and for many action but the intended one is to share)

Right, makes sense. Am still a bit on the fence cause it will create a difference with the web interface. What do you think about the icon to take @JorisBodin @Ivansss?

immortal79:
Didn't think about the whole picture. You are right it will be more consistent if it's the same icon.
@JorisBodin also agrees with this vision.

However I think it could be a little more "iOS looking" I can maybe make it less bold.

jancborchardt:
No worries – cool! :)

> However I think it could be a little more "iOS looking" I can maybe make it less bold.

Yep, that sounds perfect! The Material Design icon is less bold, maybe that’s fitting?
image
https://material.io/resources/icons/?search=share&icon=share&style=baseline

JorisBodin:
@jancborchardt 
I prefer to use iOS type icons rather than material icons on iOS :D
For exemple in iOS Glyph (The most used standard on iOS today):
image
https://icons8.com/icons/set/share

marinofaggiana:
> > Note: the icons used in other views as for example the activity must be a vector pdf with large dim. But is not a problem because I check that before the merge
> 
> I can export with larger dim but why do that ? As PDF is vector based this shouldn't make a difference ?

Because do not exists 3 icon (1x 2x 3x) + BIG for example used ad background when  the view is empty, but exists only a pdf vector "All Universal" of x cm.

jancborchardt:
> I prefer to use iOS type icons rather than material icons on iOS :D

Fine with me – the icons are almost the same anyway because it’s a 3rd party site :)
share-24px

immortal79:
As suggested by @jancborchardt , the material icon is fine when the right size is applied ;)
I also changed the dimension of the PDF icons to larger dimensions as requested.

The share icon now looks like this:
Simulator Screen Shot - iPhone X - iOS 13 - 2020-05-28 at 13 42 16

PR #1280, opened 2020-05-29
Fix for navbar behavior

immortal79:
It's a lot like the previous commit which was reverted but I added some fixes which should prevent the behaviours with the overlapping tableview/header you described.

If you find some cases where the sort header is not correct tell me how to reproduce the bug so that I can fix it.


marinofaggiana:
@immortal79 change the object from UIViewController to UITableViewController is so important for ?

immortal79:
As it's the way Apple "intended" it, it resolves a lot of bugs and layout issues with large titles / table header and search controller.

immortal79:
Also fixes #1307 ^^

marinofaggiana:
ahha

PR #1416, opened 2020-09-10
l10n: Changing the syntax of a sentence

Valdnet:


marinofaggiana:
Thanks, added manually and without final "." :-)

Valdnet:
@marinofaggiana In my opinion, there should also be a period at the end of the last sentence, as these are complex sentences. Note that this information consists of three sentences 😉.

PR #1468, opened 2020-10-29
Fix poor scanning quality

zero0cool0:
I believe commit cb8282cb5f483061c8645d7b3cec1c783f4ba551 introduced a regression where document scanning now produces really poor images, mainly due to down-scaling every scan to the allowed (595.2px x 841.8px). In practice, this produces blurry PDFs and JPGs.
This PR is trying to revert some of the previous commit by re-introducing a scale factor (similar to the original dpi), allowing for more image resolution data to be returned.


zero0cool0:
@marinofaggiana could you please comment as to why this PR was closed and what I can do to improve?

marinofaggiana:
Yes, because in "Develop" is already fixed :-)

PR #1492, opened 2020-11-28
typo fix

johkoenig:
Fixing typos: Adding spaces after the commas and removing the word "and", since this string is used in all i18n versions.


marinofaggiana:
Thanks @johkoenig

PR #1493, opened 2020-11-28
change default datetime filename

### LINKED ISSUE #359: use a better default naming scheme for instant upload

johkoenig:
fixes #359


marinofaggiana:
fixed.

PR #1540, opened 2021-02-09
change the itemForPage from 100 to 150

ZhenyuanBo:
I think it would be better to increase number of items for each page to be increased from 100 to 150.


marinofaggiana:
Hi @ZhenyuanBo , thanks for your PR but the memory under FileProvider is very very very limited and unpredictable, increase the item is not a good idea ... why do you want increase it ?

PR #1557, opened 2021-03-22
Improve issue template

thgoebel:
Things I changed:

- Removed the white space at the end of actual+expected behavour
- Move Steps to Reproduce up to the top. I think this improves the flow: first go through the steps, then here's what happens. The server and Android repo also have it in this order.
- Add sections for screenshots and logs
- Move the entire version stuff into a single section, so it takes up less space.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,23 +1,41 @@
-### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
+### Steps to reproduce
 
-### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
+1.
+2.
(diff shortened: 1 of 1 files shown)


thgoebel:
Regarding the server config section:

I wonder though whether the server OS, database and PHP version are really needed. Can't the PHP version for example be deduced from the Nextcloud server version?

Of course, I can't judge since I don't do iOS development (i.e. I don't have a good overview of what kind of issues you get, and what info is usually required to reproduce them). But I noticed that [Android](https://github.com/nextcloud/android/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) only requests the NC server version and whether there is a reverse proxy (which is where many issues with self-hosted installs come from, as far as I understand it).

marinofaggiana:
Thanks a lot @thgoebel

PR #1622, opened 2021-06-18
Various translations [nl]

rayzilt:
Various translations in 'nl' language file.


rayzilt:
Translation is not needed as done by transifex

PR #1644, opened 2021-07-15
Sharing dialog improvement 

TSImc:
Improvement of sharing dialog for iOS client
This is Similar to the PR done for Android  - [https://github.com/nextcloud/android/pull/8598/commits]

**Overview**

- Improve sharing process for Internal and External Share
- Popup sharing option menu changed to bottom sheet actions
- Added a new button for fast viewing and updating the permissions

**Screen shots** 

Screenshot 2021-07-15 at 4 51 38 PM
Screenshot 2021-07-15 at 4 51 26 PM
Screenshot 2021-07-15 at 4 51 15 PM
Screenshot 2021-07-15 at 4 51 05 PM


marinofaggiana:
@tschiu  oh good, very good, but is better as "source code" &  for "commits into," use the nextcloud:develop because the master is already old and we have make modify in the Share views

PR #1698, opened 2021-09-29
Handle "WEB" notification action.

### LINKED ISSUE #1697: Handle "WEB" actions correctly in notifications
At the moment every [notification action](https://github.com/nextcloud/server/blob/master/lib/public/Notification/IAction.php#L37-L53) is passed to [setNotification()](https://github.com/nextcloud/ios/blob/master/iOSClient/Notification/NCNotification.swift#L274)
Since "WEB" is not a valid http request method a 405 errors is thrown.

Reference: https://help.nextcloud.com/t/error-405-method-not-allowed/123152

Ivansss:
Open the browser when notification action is of type "WEB"

Fix #1697


nickvergessen (approved):
Can't test, but makes sense from the looks of it

PR #1710, opened 2021-10-11
Fix user display

### LINKED ISSUE #1703: Coordinate displaying avatars
Test Admin -> TA
Tobias Kaminsky -> TK
Admin -> A
Test Server Admin -> TS
email@server.com -> E

Use first char of first two words:

```
$firstTwoLetters = array_map(function ($namePart) {
			return mb_strtoupper(mb_substr($namePart, 0, 1), 'UTF-8');
		}, explode(' ', $displayName, 2));
```

thisIsTheFoxe:
Fix #1703
https://github.com/nextcloud/server/issues/22434

The main problem was `NCCommunication.shared.downloadAvatar(user: user ...)` instead of using the `userId`, causing a 404 error. If the user didn't set a custom picture the server returns the *initials-avatar* by default.
If no image is available (e.g. no server connection, server error) create an initials-avatar locally. If display name is empty fallback to user icon.

Also, on the Files tab have the display name as title, if no alias is set (Idea was from Joas iirc).

<hr/>

Preview:
IMG_59AA145C70AC-1

### DIFF
--- a/iOSClient/NCGlobal.swift
+++ b/iOSClient/NCGlobal.swift
@@ -42,9 +42,9 @@ class NCGlobal: NSObject {
         if (!matches.isEmpty) {
             // Already a md5 hash?
             // done, use as is.
-            hash = loweUserName
+            hash = lowerUsername
         } else {
-            hash = loweUserName.md5()
+            hash = lowerUsername.md5()
         }
 
         hash = hash.replacingOccurrences(of: "[^0-9a-f]", with: "", options: .regularExpression)


tobiasKaminsky:
>  instead of using the `userId`

So this uses always userId? Also for generating the "fallback" 2 char name?
userId is the internal id, which might be for LDAP something like d083fa0e-c26a-45b0-94da-bc204c6a7cd4.
Therefore "display name" needs to be used, as this is the readable name, e.g. "Tobias Kaminsky", whereas userId is "tobias" on c.nc.c

PR #1711, opened 2021-10-12
add CI

thisIsTheFoxe:
Add CI for:
- xcodebuild
  - cartage bootstrap
- SwiftLint

## Build
The build process uses the default GitHub Actions Xcode version with macOS 11. It uses [actions/cache](https://github.com/actions/cache) to cache Carthage dependencies (if *Cartfile.resolved* file hasn't changed). Then run `xcodebuild clean build` for each target (specified in _xcode.yml_) plus run the tests for main Nextcloud app.

## Lint
As this is a big project with **A LOT** of listing issues. I propose the following:
1. Use `SwiftLint --fix` to fix simple issues (like formatting & whitespaces) throughout the project (included in this PR)
2. `exclude` any files that still contain SwiftLint warnings or errors.
3. (ideally) when a file is edited, remove the file from the quarantined list in `.swiftlint.yml` and fix all lint issues in that file.


Ivansss:
> Talk has the DCO ? (NCCommunication yes)

Yes, Talk iOS has DCO.

tobiasKaminsky:
> As this is a big project with **A LOT** of listing issues. I propose the following:
> 
>     1.

We use https://github.com/nextcloud/android/blob/master/scripts/analysis/lint-up.rb on Android.
This runs lint, compares result to stored result and if it is
- less -> updates, then re-run
- same -> green CI
- more -> CI fails

This way the project will not increase lint error/warnings, but can only stay at the same level, or getting better.

tobiasKaminsky:
@thisIsTheFoxe can you extract xcode building process, so that we can merge/use this soon?
Whole lint can then be done later (or via my proposed lint-up script)

thisIsTheFoxe:
> We use https://github.com/nextcloud/android/blob/master/scripts/analysis/lint-up.rb on Android.
> 
> This way the project will not increase lint error/warnings, but can only stay at the same level, or getting better.

Thanks for sharing @tobiasKaminsky, that's interesting.. and I've seen that mostly for code coverage CIs..

However, any linting *errors* will cause the CI to fail anyway. Which is good because an error usually means that the code check ran into a severe issue that might case bugs or crashes. So, if you meant to use something like that _instead_ of my approach (ignoring files) I don't think that'd work because a) you'd still have to silence the errors somehow, and b) seeing thousands of lint warnings might be somewhat desensitising.
If your goal is to just have a report on linting issues and maybe even visualise them let me know, bc I haven't looked into that much yet.

Ivansss (approved):
LGTM 👍

marinofaggiana (approved):
I can't look into it, it would take me too long that I don't have, so I approve without look

alazarevic5 (approved):
Approved. Code style 👍🏼

PR #1742, opened 2021-11-08
Fix keep both conflict action

thisIsTheFoxe:
Rename file metadata and then move file so the new filename matches the file saved on disk.
This can then be used to upload the file etc. Otherwise the `upload` function will fail bc the file cannot be found.

### DIFF
--- a/iOSClient/Main/Create cloud/NCCreateFormUploadConflict.swift
+++ b/iOSClient/Main/Create cloud/NCCreateFormUploadConflict.swift
@@ -181,7 +181,8 @@ extension NCCreateFormUploadConflictDelegate {
             for metadata in self.metadatasUploadInConflict {
                 
                 let fileNameMOV = (metadata.fileNameView as NSString).deletingPathExtension + ".mov"
-                let newFileName = NCUtilityFileSystem.shared.createFileName(metadata.fileNameView, serverUrl: metadata.serverUrl, account: metadata.account)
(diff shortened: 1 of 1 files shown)


marinofaggiana:
I must check ... I don't remember what happens and why use fileNameMOV (Live Photo ... ?)

marinofaggiana:
@thisIsTheFoxe as imagined... don't works with the LivePhoto image ( image + mov), why do you have totally  removed the fileNameMOV ?

thisIsTheFoxe:
I'm unsure of what exactly it does. And as far as I can see `self.metadatasMOV` is not being modified so the nested for loop doesn't make much sense to me and is very inefficiently doing the exact same thing multiple times.
Also initially I thought the *.mov* extension might just be an aid for renaming the file.. it can be quite confusing, sometimes..

marinofaggiana:
No, totally incorrect: 

let metadataMOV = self.metadatasMOV.first(where: { $0.ocId == fileNameMOV }),  $0.ocId == fileNameMOV ? will be always false

-1- upload a Live Photo with the "Maintain original filename" ON

Simulator Screen Shot - iPhone 13 - 2021-11-09 at 15 05 45

- 2 - upload the same Photo so you can look the two files, one the image and one the mov part, the file in conflict are two and not one only.

Before create a new PR please try if works.

thanks

PR #1744, opened 2021-11-09
Fix duplicate shares

marinofaggiana:
In tab "More" -> Shares, if one 'file' is shared with multiple users it's shown multiple times.

Signed-off-by: marinofaggiana <marino@marinofaggiana.com>


thisIsTheFoxe (approved):
Looks good, thanks.
Possible optimisation: use a set to keep track of which share-`ocId`s are shown instead of checking with `.contains(where:)`.

marinofaggiana:
Completely agree but for low number is more simple use a simple contains 😄

PR #1745, opened 2021-11-09
Added in some view the IHProgressHUD

marinofaggiana:
Some view needs a better HUD with the progress, as the upload files in the `share extension`, so I have found this library that is not perfect (I don't like the `singleton`) but is a valid for the extension because can used the `view` and not only the `appDelegate.window?.rootViewController?`

### DIFF
--- a/Share/NCShareExtension.swift
+++ b/Share/NCShareExtension.swift
@@ -458,7 +458,7 @@ class NCShareExtension: UIViewController, NCListCellDelegate, NCEmptyDataSetDele
             
             IHProgressHUD.showSuccesswithStatus(NSLocalizedString("_success_", comment: ""))
 
-            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
                 self.extensionContext?.completeRequest(returningItems: self.extensionContext?.inputItems, completionHandler: nil)
             }
         }
--- a/Share/NCShareExtension.swift
+++ b/Share/NCShareExtension.swift
@@ -23,6 +23,7 @@
 
 import UIKit
 import NCCommunication
+import IHProgressHUD
 
 class NCShareExtension: UIViewController, NCListCellDelegate, NCEmptyDataSetDelegate, NCRenameFileDelegate, NCAccountRequestDelegate {
     
@@ -64,6 +65,7 @@ class NCShareExtension: UIViewController, NCListCellDelegate, NCEmptyDataSetDele
     private let refreshControl = UIRefreshControl()
     private var activeAccount: tableAccount!
     private let chunckSize = CCUtility.getChunkSize() * 1000000
+    private var fileNameInUpload: String?
     
     // MARK: - View Life Cycle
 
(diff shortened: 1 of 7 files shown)


Share/NCShareExtension.swift
```diff
@@ -65,6 +66,9 @@ class NCShareExtension: UIViewController, NCListCellDelegate, NCEmptyDataSetDele
     private var activeAccount: tableAccount!
     private let chunckSize = CCUtility.getChunkSize() * 1000000
     
+    private var numberFilesName: Int = 0
```
thisIsTheFoxe, line 69: Is this being used for something else or is there another reason you can't just do:
```suggestion
    private var numberFilesName: Int { filesName.count }
```

Share/NCShareExtension.swift
```diff
@@ -412,28 +435,33 @@ class NCShareExtension: UIViewController, NCListCellDelegate, NCEmptyDataSetDele
...
                     }
                 }
             }
         } else {
-            extensionContext?.completeRequest(returningItems: extensionContext?.inputItems, completionHandler: nil)
-            return
+            
+            IHProgressHUD.showSuccesswithStatus(NSLocalizedString("_success_", comment: ""))
+
+            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
```
thisIsTheFoxe, line 462: 2s seems a bit arbitrary and personally a bit too long for me. If everything succeeded it should just dismiss right away - imo.

iOSClient/Main/NCFunctionCenter.swift
```diff
@@ -352,34 +353,34 @@ import Queuer
         }
         
         if CCUtility.fileProviderStorageExists(metadata.ocId, fileNameView: metadata.fileNameView) && CCUtility.fileProviderStorageExists(metadataMOV.ocId, fileNameView: metadataMOV.fileNameView) {
-            saveLivePhotoToDisk(metadata: metadata, metadataMov: metadataMOV, progressView: nil, viewActivity: self.appDelegate.window?.rootViewController?.view)
+            saveLivePhotoToDisk(metadata: metadata, metadataMov: metadataMOV)
         }
     }
     
-    func saveLivePhotoToDisk(metadata: tableMetadata, metadataMov: tableMetadata, progressView: UIProgressView?, viewActivity: UIView?) {
+    func saveLivePhotoToDisk(metadata: tableMetadata, metadataMov: tableMetadata) {
```
thisIsTheFoxe, line 360: Again it seems a bit random why the change is made here and `UIProgressView` is still being used in other parts of the app..?

thisIsTheFoxe (changes requested):
Thank, tho I'm not yet entirely sure why the change is necessary. 
Anyway, a few other UX concerns:
- using both `IHProgressHUD` and `UIProgressView` might make the app feel inconsistent
- when displaying the `IHProgressHUD` the entire view becomes unresponsive. Which feels unintuitive to me. If that's intended, may I suggest darkening the rest like system alerts do?

Share/NCShareExtension.swift
marinofaggiana: true ... but some App(s) displays the final message (success) for a few seconds for comforts the user, other close immediataly the view .... it always depends on the point of view, ...

marinofaggiana:
> Thank, tho I'm not yet entirely sure why the change is necessary. Anyway, a few other UX concerns:
> 
> * using both `IHProgressHUD` and `UIProgressView` might make the app feel inconsistent
> * when displaying the `IHProgressHUD` the entire view becomes unresponsive. Which feels unintuitive to me. If that's intended, may I suggest darkening the rest like system alerts do?

The hud is different from the UIProgressView and the use of one does not exclude the other as well as the unresponsive ... depends on where and how it is used.

Share/NCShareExtension.swift
marinofaggiana: @Ivansss @alazarevic5 what's your opinion...

marinofaggiana: mmm nope ... @thisIsTheFoxe 0, 1 or 2 sec. ?

iOSClient/Main/NCFunctionCenter.swift
marinofaggiana: It's present in Viewer image for the progress download image

Share/NCShareExtension.swift
marinofaggiana: 1 for me the user must see the "success""

marinofaggiana:
Ok we can try it

PR #1749, opened 2021-11-10
Content presenter

marinofaggiana:
Improvements of the `messageNotification` function with remove of old parameter `forced` and added the new parameters `priority` & `dropEnqueuedEntries` for managed the queue of message


iOSClient/Utility/NCContentPresenter.swift
```diff
@@ -60,63 +60,56 @@ class NCContentPresenter: NSObject {
...
-                lastErrorCode = errorCode
-            }
-        }
+        // No notification message for:
+        if errorCode == -999 { return }         // Cancelled transfer
+        else if errorCode == 200 { return }     // Transfer stopped
+        else if errorCode == 207 { return }     // WebDAV multistatus
+        //else if errorCode == 423 { return }     // WebDAV locked
+        //else if errorCode == -1001 { return }   // Time out
+        //else if errorCode == -1005 { return }   // Connection lost
```
thisIsTheFoxe, line 77: Are those still needed? if not remove to avoid clutter..

iOSClient/Utility/NCContentPresenter.swift
```diff
@@ -60,63 +60,56 @@ class NCContentPresenter: NSObject {
...
+        else if errorCode == NCGlobal.shared.errorNoError && type == messageType.error { return }
         
         DispatchQueue.main.async {
             switch errorCode {
             case Int(CFNetworkErrors.cfurlErrorNotConnectedToInternet.rawValue):
                 let image = UIImage(named: "networkInProgress")!.image(color: .white, size: 20)
-                self.noteTop(text: NSLocalizedString(title, comment: ""), image: image, color: .lightGray, delay: delay, name: "\(errorCode)")
+                self.noteTop(text: NSLocalizedString(title, comment: ""), image: image, color: .lightGray, delay: delay, priority: .max)
```
thisIsTheFoxe, line 84: Why is here `.max` hardcoded? Why not use the `priority` parameter? If I see it correctly, the parameter is not used elsewhere. If that's on purpose remove the `priority` parameter from the function.

iOSClient/Utility/NCContentPresenter.swift
```diff
@@ -60,63 +60,56 @@ class NCContentPresenter: NSObject {
...
             //case Int(kOCErrorServerUnauthorized), Int(kOCErrorServerForbidden):
             //    break
```
thisIsTheFoxe, line 86: More unneeded comments..

iOSClient/Utility/NCContentPresenter.swift
```diff
@@ -60,63 +60,56 @@ class NCContentPresenter: NSObject {
...
         }
     }
     
     //MARK: - Flat message
     
-    @objc func flatTop(title: String, description: String, delay: TimeInterval, imageName: String?, type: messageType, name: String?) {
+    private func flatTop(title: String, description: String, delay: TimeInterval, imageName: String?, type: messageType, priority: EKAttributes.Precedence.Priority = .normal, dropEnqueuedEntries: Bool = false) {
      
-        if name != nil && SwiftEntryKit.isCurrentlyDisplaying(entryNamed: name) { return }
+        if SwiftEntryKit.isCurrentlyDisplaying(entryNamed: title+description) { return }
```
thisIsTheFoxe, line 100:
```suggestion
        if SwiftEntryKit.isCurrentlyDisplaying(entryNamed: title + description) { return }
```
Nitpick 😄

iOSClient/Utility/NCContentPresenter.swift
```diff
@@ -60,63 +60,56 @@ class NCContentPresenter: NSObject {
...
         
         var attributes = EKAttributes.topFloat
         var image: UIImage?
         
         attributes.windowLevel = .normal
         attributes.displayDuration = delay
-        attributes.name = name
+        attributes.name = title+description
```
thisIsTheFoxe, line 107:
```suggestion
        attributes.name = title + description
```

thisIsTheFoxe (changes requested):
Mostly just nits..
Except for line 84, please check if this is correct..

iOSClient/Utility/NCContentPresenter.swift
marinofaggiana: Because the message notification of NotConnectedToInternet required always the MAX priority

thisIsTheFoxe: Ok, in that case please remove the `priority` paramter from `messageNotification(...)`.

iOSClient/Utility/NCContentPresenter.swift
marinofaggiana: yes the -999 is not a error message as 200 and 207

iOSClient/Utility/NCContentPresenter.swift
marinofaggiana: Please ?

iOSClient/Utility/NCContentPresenter.swift
thisIsTheFoxe: I was talking about the commented lines (75-77). If they are needed for clarity add it to the function's documentation. Otherwise please remove them.

iOSClient/Utility/NCContentPresenter.swift
thisIsTheFoxe: The `priority` parameter is unused in [`messageNotification`](https://github.com/nextcloud/ios/blob/2e1bd0ac0165c4dce039d626d6cc6acc77964929/iOSClient/Utility/NCContentPresenter.swift#L69)

marinofaggiana: no, it's used and it's important for some important notification and future notification

thisIsTheFoxe: It is not used in that function and causes confusion, because it does nothing. Please remove it.

iOSClient/Utility/NCContentPresenter.swift
```diff
@@ -60,63 +60,51 @@ class NCContentPresenter: NSObject {
...
-    @objc private var lastErrorCode: Int = 0
-
     //MARK: - Message
     
-    @objc func messageNotification(_ title: String, description: String?, delay: TimeInterval, type: messageType, errorCode: Int, forced: Bool = false) {
+    @objc func messageNotification(_ title: String, description: String?, delay: TimeInterval, type: messageType, errorCode: Int) {
+        messageNotification(title, description: description, delay: delay, type: type, errorCode: errorCode, priority: .normal, dropEnqueuedEntries: false)
+    }
+    
+    func messageNotification(_ title: String, description: String?, delay: TimeInterval, type: messageType, errorCode: Int, priority: EKAttributes.Precedence.Priority = .normal, dropEnqueuedEntries: Bool = false) {
```
thisIsTheFoxe, line 69: I still don't see the use of this parameter. It is not used in the function. If it is needed for future reference please add it as documentation.
```suggestion
    func messageNotification(_ title: String, description: String?, delay: TimeInterval, type: messageType, errorCode: Int, dropEnqueuedEntries: Bool = false) {
```

thisIsTheFoxe (changes requested):

iOSClient/Utility/NCContentPresenter.swift
marinofaggiana: https://github.com/nextcloud/ios/blob/3508a96adf50aa30f2f9c5c8998cd05eb7a18b4a/iOSClient/Networking/NCNetworkingE2EE.swift#L262

marinofaggiana: have you look the callers in this PR ? or I didn't understand :D

thisIsTheFoxe: Thanks for the example, but that is a) not the `messageNotification` function and b) the caller, not the implementation.
Please check the [entire function](https://github.com/nextcloud/ios/blob/3508a96adf50aa30f2f9c5c8998cd05eb7a18b4a/iOSClient/Utility/NCContentPresenter.swift#L69-L89) again. The `priority` parameter is not used.

marinofaggiana: uuu sorry my error

marinofaggiana: ja ja ... my error, I Correct.

marinofaggiana: https://github.com/nextcloud/ios/blob/3508a96adf50aa30f2f9c5c8998cd05eb7a18b4a/iOSClient/Networking/NCNetworking.swift#L208

thisIsTheFoxe: Yes, but this is the same as:
```swift
NCContentPresenter.shared.messageNotification("_error_", description: "_error_creation_file_", [...], priority: .normal)
```
There's no difference in the implementation and execution.

marinofaggiana: very strange I remember putting it on ... look now

thisIsTheFoxe (approved):
Lgtm
Please make sure the DCO is passing.

PR #1751, opened 2021-11-11
Fix/activities

thisIsTheFoxe:
Viewing Activities in "More" tab have not metadata and just display all activities.
Also, can't make general comments without a fileId.
Also, fix large navigation title.


marinofaggiana:
Please, removed the Lint "Expected — Waiting for status to be reported"

marinofaggiana (approved):

PR #1757, opened 2021-11-15
Fix Activities view

### LINKED ISSUE #1752: New Activity view + Comment
1. This activity indicator seems be too much 

Simulator Screen Shot - iPhone 13 Pro - 2021-11-11 at 19 06 28

2. How many time is called the API during the refresh .. ?

Schermata 2021-11-11 alle 19 16 36

3. The "Cancel" seem be useless

Simulator Screen Shot - iPhone 13 Pro - 2021-11-11 at 19 05 10

thisIsTheFoxe:
- Fix #1752 

TODO: 
- [x] DCO

### DIFF
--- a/iOSClient/Activity/NCActivity.swift
+++ b/iOSClient/Activity/NCActivity.swift
@@ -132,8 +132,7 @@ class NCActivity: UIViewController {
     // MARK: - NotificationCenter
 
     @objc func initialize() {
-        loadActivity(idActivity: 0)
-        loadComments()
+        fetchAll(initial: true)
     }
     
     @objc func changeTheming() {
@@ -353,10 +352,9 @@ extension NCActivity: UIScrollViewDelegate {
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
         guard
             scrollView.contentOffset.y > 50,
-            scrollView.contentSize.height - scrollView.frame.height - scrollView.contentOffset.y < -50,
-            let activity = allItems.compactMap({ $0 as? tableActivity }).last
+            scrollView.contentSize.height - scrollView.frame.height - scrollView.contentOffset.y < -50
         else { return }
-        loadActivity(idActivity: activity.objectId)
+        fetchAll(initial: false)
     }
 }
 
@@ -364,6 +362,31 @@ extension NCActivity: UIScrollViewDelegate {
 
 extension NCActivity {
 
+    func fetchAll(initial: Bool) {
+        guard canFetchActivity else { return }
+        self.canFetchActivity = false
+
+        let height = self.tabBarController?.tabBar.frame.size.height ?? 0
+        NCUtility.shared.startActivityIndicator(backgroundView: self.view, blurEffect: false, bottom: height + 50, style: .gray)
+
+        let dispatchGroup = DispatchGroup()
+        loadComments(disptachGroup: dispatchGroup)
+        if initial {
+            loadActivity(idActivity: 0, disptachGroup: dispatchGroup)
+        } else if let activity = allItems.compactMap({ $0 as? tableActivity }).last {
+            loadActivity(idActivity: activity.objectId, disptachGroup: dispatchGroup)
+        }
+        
+        dispatchGroup.notify(queue: .main) {
+            self.loadDataSource()
+            NCUtility.shared.stopActivityIndicator()
+            // otherwise is triggered again
+            DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+                self.canFetchActivity = true
+            }
+        }
+    }
+    
     func loadDataSource() {
 
         var newItems = [DateCompareable]()
@@ -385,29 +408,27 @@ extension NCActivity {
         self.tableView.reloadData()
     }
     
-    func loadComments() {
-        guard let metadata = metadata else { return }
+    func loadComments(disptachGroup: DispatchGroup? = nil) {
+        guard showComments, let metadata = metadata else { return }
+        disptachGroup?.enter()
+
         NCCommunication.shared.getComments(fileId: metadata.fileId) { (account, comments, errorCode, errorDescription) in
             if errorCode == 0, let comments = comments {
                 NCManageDatabase.shared.addComments(comments, account: metadata.account, objectId: metadata.fileId)
-                self.loadDataSource()
+            } else if errorCode != NCGlobal.shared.errorResourceNotFound {
+                NCContentPresenter.shared.messageNotification("_share_", description: errorDescription, delay: NCGlobal.shared.dismissAfterSecond, type: NCContentPresenter.messageType.error, errorCode: errorCode)
+            }
+
+            if let disptachGroup = disptachGroup {
+                disptachGroup.leave()
             } else {
-                if errorCode != NCGlobal.shared.errorResourceNotFound {
-                    NCContentPresenter.shared.messageNotification("_share_", description: errorDescription, delay: NCGlobal.shared.dismissAfterSecond, type: NCContentPresenter.messageType.error, errorCode: errorCode)
-                }
+                self.loadDataSource()
             }
         }
     }
     
-    func loadActivity(idActivity: Int) {
-
-        guard canFetchActivity else { return }
-        canFetchActivity = false
-
-        if idActivity > 0 {
-            let height = self.tabBarController?.tabBar.frame.size.height ?? 0
-            NCUtility.shared.startActivityIndicator(backgroundView: self.view, blurEffect: false, bottom: height + 50, style: .gray)
-        }
+    func loadActivity(idActivity: Int, disptachGroup: DispatchGroup? = nil) {
+        disptachGroup?.enter()
 
         NCCommunication.shared.getActivity(
             since: idActivity,
@@ -418,13 +439,12 @@ extension NCActivity {
                 
                 if errorCode == 0 && account == self.appDelegate.account {
                     NCManageDatabase.shared.addActivity(activities, account: account)
-                    self.loadDataSource()
                 }
                 
-                NCUtility.shared.stopActivityIndicator()
-                
-                DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
-                    self.canFetchActivity = errorCode != NCGlobal.shared.errorNotModified
+                if let disptachGroup = disptachGroup {
+                    disptachGroup.leave()
+                } else {
+                    self.loadDataSource()
                 }
             }
     }
--- a/iOSClient/Activity/NCActivity.swift
+++ b/iOSClient/Activity/NCActivity.swift
@@ -132,6 +132,7 @@ class NCActivity: UIViewController {
     // MARK: - NotificationCenter
 
     @objc func initialize() {
+        loadDataSource()
         fetchAll(isInitial: true)
     }
     
@@ -372,7 +373,7 @@ extension NCActivity {
         let dispatchGroup = DispatchGroup()
         loadComments(disptachGroup: dispatchGroup)
         if !isInitial, let activity = allItems.compactMap({ $0 as? tableActivity }).last {
-            loadActivity(idActivity: activity.objectId, disptachGroup: dispatchGroup)
+            loadActivity(idActivity: activity.idActivity, disptachGroup: dispatchGroup)
         } else {
             loadActivity(idActivity: 0, disptachGroup: dispatchGroup)
         }
--- a/iOSClient/Activity/NCActivity.swift
+++ b/iOSClient/Activity/NCActivity.swift
@@ -25,7 +25,6 @@
 import UIKit
 import SwiftRichString
 import NCCommunication
-import RealmSwift
 
 class NCActivity: UIViewController {
 
@@ -485,9 +484,9 @@ extension NCActivity: NCShareCommentsCellDelegate {
     func tapMenu(with tableComments: tableComments?, sender: Any) {
         toggleMenu(with: tableComments)
     }
-    
+
     func toggleMenu(with tableComments: tableComments?) {
-        
+
         let menuViewController = UIStoryboard.init(name: "NCMenu", bundle: nil).instantiateInitialViewController() as! NCMenu
         var actions = [NCMenuAction]()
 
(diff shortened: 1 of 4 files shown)


iOSClient/Activity/NCActivity.swift
```diff
@@ -351,69 +350,86 @@ extension NCActivity: UITableViewDataSource {
...
+            if let disptachGroup = disptachGroup {
+                disptachGroup.leave()
+            } else {
+                self.loadDataSource()
+            }
         }
-        
+    }
+    
+    func loadActivity(idActivity: Int, disptachGroup: DispatchGroup? = nil) {
```
marinofaggiana, line 430: loadActivity is called always with disptachGroup(), so this arguments can't be optional (?)

iOSClient/Activity/NCActivity.swift
```diff
@@ -351,69 +350,86 @@ extension NCActivity: UITableViewDataSource {
...
+        disptachGroup?.enter()
```
marinofaggiana, line 431: remove ?

iOSClient/Activity/NCActivity.swift
```diff
@@ -425,15 +441,11 @@ extension NCActivity {
                     NCManageDatabase.shared.addActivity(activities, account: account)
                 }
                 
-                NCUtility.shared.stopActivityIndicator()
-                
-                if errorCode == NCGlobal.shared.errorNotModified {
-                    self.canFetchActivity = false
+                if let disptachGroup = disptachGroup {
```
marinofaggiana, line 444: replace with disptachGroup.leave()

iOSClient/Activity/NCActivity.swift
thisIsTheFoxe: Correct 👍

marinofaggiana:
The load of old activity don't works, this capture view the last activity load after don't see anything but the activity via web site are much older:

Simulator Screen Shot - iPhone 13 - 2021-11-16 at 09 22 25

web:

Schermata 2021-11-16 alle 09 26 08

marinofaggiana:
Note optimization: 

1. when start view immediately the item stored. (#now I must wait the loading ... )

2. when load the view reload always the first 200 item, it would be better load 1 item and compare if exists in the database, if yes stop otherwise load the other items

what's do you think

marinofaggiana:
@thisIsTheFoxe You can look here for API detail (improvements) :

https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md

thisIsTheFoxe:
> 2. when load the view reload always the first 200 item, it would be better load 1 item and compare if exists in the database, if yes stop otherwise load the other items

Problem: I edit a file and check the details (load newest activity). Next I check all Activities (in "More" tab). Now, the first activity is loaded, skip loading new activities. But now the rest, everything in between my edit and the last activity that was saved will *never* be loaded.

marinofaggiana:
> > 1. when load the view reload always the first 200 item, it would be better load 1 item and compare if exists in the database, if yes stop otherwise load the other items
> 
> Problem: I edit a file and check the details (load newest activity). Next I check all Activities (in "More" tab). Now, the first activity is loaded, skip loading new activities. But now the rest, everything in between my edit and the last activity that was saved will _never_ be loaded, right?

load newest activity --> will be load from newest activity TO last activity stored in our DB -> filter for METADATA , correct ?

thisIsTheFoxe:
> > > 1. when load the view reload always the first 200 item, it would be better load 1 item and compare if exists in the database, if yes stop otherwise load the other items
> > 
> > 
> > Problem: I edit a file and check the details (load newest activity). Next I check all Activities (in "More" tab). Now, the first activity is loaded, skip loading new activities. But now the rest, everything in between my edit and the last activity that was saved will _never_ be loaded, right?
> 
> load newest activity --> will be load from newest activity TO last activity stored in our DB -> filter for METADATA , correct ?

First, not sure what you mean, I don't think it's possible it's possible to specify a range like that in the API. It'll always have to load all the newest activities.

Filtering is the problem. Because I don't think we should change the behaviour of that. If I view file activities it should only load file activities and not everything. Otherwise it might take a long time to load recent activities for that specific file. 
But that also means that there is no way of knowing whether all recent activities have been loaded, if the most recent activities match.

marinofaggiana:
> First, not sure what you mean, I don't think it's possible it's possible to specify a range like that in the API. It'll always have to load all the newest activities.

Sure that is possible: 

GET /ocs/v2.php/apps/activity/api/v2/activity 
Parameters
since: nothing
limit: 1
return example ID: 1224

Now get your last ID from DB example 1220:

GET /ocs/v2.php/apps/activity/api/v2/activity 
Parameters
since 1224
limit 4 (1224-1220) 

works for you ?

> 
> Filtering is the problem. Because I don't think we should change the behaviour of that. If I view file activities it should only load file activities and not everything. Otherwise it might take a long time to load recent activities for that specific file. But that also means that there is no way of knowing whether all recent activities have been loaded, if the most recent activities match.

I don't understand...

thisIsTheFoxe:
> Now get your last ID from DB example 1220:
GET /ocs/v2.php/apps/activity/api/v2/activity
Parameters
since 1224
limit 4 (1224-1220)

Yes, but that only means 1220 is the most recent. It doesn't guarantee that from 1220 all activities are actually loaded.
 
Problem: You edit some file (activityId: 1219). Next, I edit another file and check the details (load newest activity, filtered for file; activityId: 1220). Next, I check all Activities (in "More" tab). Now, the first activity in the DB has the id 1220 and the last say, 1000. So I skip reloading and never load activity 1219.

marinofaggiana:
> > Now get your last ID from DB example 1220:
> > GET /ocs/v2.php/apps/activity/api/v2/activity
> > Parameters
> > since 1224
> > limit 4 (1224-1220)
> 
> Yes, but that only means 1220 is the most recent. It doesn't guarantee that from 1220 all activities are actually loaded.
> 
> Problem: You edit some file (activityId: 1219). Next, I edit another file and check the details (load newest activity, filtered for file; activityId: 1220). Next, I check all Activities (in "More" tab). Now, the first activity in the DB has the id 1220. So I load 1224-1220, but **never** load 1219.

because you use the filtered for file object_type in the (ACTIVITY+COMMENTS) ... so is correct but you don't have to memorize it on DB  the newest activity (it's however wrong, looking at it today,load every time 200 items).

Solution ?

- Modify DB for this two possibility (from file or not, another column maybe activityId_files)
- or don't memorized the fetch from object_type [no good ... but possible]
- or don'u use the object_type (I don't think you generate so many records / user ... #now load 200 item for nothing)

...

marinofaggiana:
Your last push:

Schermata 2021-11-16 alle 13 59 11

where:

403 Forbidden | The offset activity belongs to a different user

thisIsTheFoxe:
> Modify DB for this two possibility (from file or not, another column maybe activityId_files)

That would mean saving the activities twice. Wasting space.
> or don't memorized the fetch from object_type [no good ... but possible]

Yes, not good.
>or don'u use the object_type (I don't think you generate so many records / user ... #now load 200 item for nothing)


Not exactly sure what you mean. But I think it's about that:
> Because I don't think we should change the behaviour of that. If I view file activities it should only load file activities and not everything. Otherwise it might take a long time to load recent activities for that specific file.


<hr/>

> Your last push:
> 
> Schermata 2021-11-16 alle 13 59 11
> 
> where:
> 
> 403 Forbidden | The offset activity belongs to a different user

A) the screenshot shows 304 (Not Modified), not 403. And I do get those for loading avatars, not activities:

```
[Result]: failure(Alamofire.AFError.responseValidationFailed(reason: Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code: 304)))
[Request]: GET https://cloud.nextcloud.com/index.php/avatar/jos/384
    [Headers]:
        If-None-Match: "779245e54a88b76e20b621463aa03d18"
    [Body]: None
[Response]:
    [Status Code]: 304
```

marinofaggiana:
> > Modify DB for this two possibility (from file or not, another column maybe activityId_files)
> 
> That would mean saving the activities twice. Wasting space.
> 
> > or don't memorized the fetch from object_type [no good ... but possible]
> 
> Yes, not good.
> 
> > or don'u use the object_type (I don't think you generate so many records / user ... #now load 200 item for nothing)
> 
> Not exactly sure what you mean. But I think it's about that:
> 
> > Because I don't think we should change the behaviour of that. If I view file activities it should only load file activities and not everything. Otherwise it might take a long time to load recent activities for that specific file.
> 

you propose a solution.

> > Your last push:
> > Schermata 2021-11-16 alle 13 59 11
> > where:
> > 403 Forbidden | The offset activity belongs to a different user
> 
> The screenshot shows 304 (Not Modified), not 403. And I do get those for loading avatars, not activities:
> 
> ```
> [Result]: failure(Alamofire.AFError.responseValidationFailed(reason: Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code: 304)))
> [Request]: GET https://cloud.nextcloud.com/index.php/avatar/jos/384
>     [Headers]:
>         If-None-Match: "779245e54a88b76e20b621463aa03d18"
>     [Body]: None
> [Response]:
>     [Status Code]: 304
> ```

uhh no no was from /ocs/v2.php/apps/activity/api/v2/activity/all?format=json&limit=200&previews=true&since=53334 
and 304 = end item. But something does not work ... via web I have other old items ... now I verify why.

marinofaggiana:
> > > Modify DB for this two possibility (from file or not, another column maybe activityId_files)
> > 
> > 
> > That would mean saving the activities twice. Wasting space.
> > > or don't memorized the fetch from object_type [no good ... but possible]
> > 
> > 
> > Yes, not good.
> > > or don'u use the object_type (I don't think you generate so many records / user ... #now load 200 item for nothing)
> > 
> > 
> > Not exactly sure what you mean. But I think it's about that:
> > > Because I don't think we should change the behaviour of that. If I view file activities it should only load file activities and not everything. Otherwise it might take a long time to load recent activities for that specific file.
> 
> you propose a solution.
> 
> > > Your last push:
> > > Schermata 2021-11-16 alle 13 59 11
> > > where:
> > > 403 Forbidden | The offset activity belongs to a different user
> > 
> > 
> > The screenshot shows 304 (Not Modified), not 403. And I do get those for loading avatars, not activities:
> > ```
> > [Result]: failure(Alamofire.AFError.responseValidationFailed(reason: Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code: 304)))
> > [Request]: GET https://cloud.nextcloud.com/index.php/avatar/jos/384
> >     [Headers]:
> >         If-None-Match: "779245e54a88b76e20b621463aa03d18"
> >     [Body]: None
> > [Response]:
> >     [Status Code]: 304
> > ```
> 
> uhh no no was from /ocs/v2.php/apps/activity/api/v2/activity/all?format=json&limit=200&previews=true&since=53334 and 304 = end item. But something does not work ... via web I have other old items ... now I verify why.

Find, database dirty.

but at the end 304 we can added in footer, like the server,  the:

Schermata 2021-11-16 alle 14 16 38

marinofaggiana:
> > you propose a solution.
> 
> Just keep a counter of the most recently loaded activity without filtering. I think that's the easiest and best compromise.

remember obligatory update (+1) the databaseSchemaVersion whenever you modify the DB

https://github.com/nextcloud/ios/blob/d2e337830d2b6c89e9c59a0516b17b54165840cf/iOSClient/NCGlobal.swift#L114

thisIsTheFoxe:
> > > The screenshot shows 304 (Not Modified), not 403. And I do get those for loading avatars, not activities:
> > > ```
> > > [Result]: failure(Alamofire.AFError.responseValidationFailed(reason: Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code: 304)))
> > > [Request]: GET https://cloud.nextcloud.com/index.php/avatar/jos/384
> > >     [Headers]:
> > >         If-None-Match: "779245e54a88b76e20b621463aa03d18"
> > >     [Body]: None
> > > [Response]:
> > >     [Status Code]: 304
> > > ```
> > 
> > 
> > uhh no no was from /ocs/v2.php/apps/activity/api/v2/activity/all?format=json&limit=200&previews=true&since=53334 and 304 = end item. But something does not work ... via web I have other old items ... now I verify why.
> 
> Find, database dirty.
> 
> but at the end 304 we can added in footer, like the server, the:
> 
> Schermata 2021-11-16 alle 14 16 38

Good catch. But imo that's a little out of scope for this PR. Can you please make a new issue for that?

marinofaggiana:
you can removed this unused include

https://github.com/nextcloud/ios/blob/286f97befbff496aa4bedc036ec34b876889cd56/iOSClient/Activity/NCActivity.swift#L28

iOSClient/Data/NCDatabase.swift
```diff
@@ -107,6 +108,14 @@ class tableAccount: Object {
...
 
+class tableActivityLatestId: Object {
+    @objc dynamic var account = ""
+    @objc dynamic var mostRecentlyLoadedActivityId: Int = 0
+    override static func primaryKey() -> String {
+        return "account"
+    }
+}
+
 class tableActivity: Object, DateCompareable {
```
marinofaggiana, line 119: small fixation, can you move this new `tableActivityLatestId` after the class `tableActivity` (sorted by letter)

marinofaggiana (changes requested):

iOSClient/Activity/NCActivity.swift
```diff
@@ -351,89 +351,132 @@ extension NCActivity: UITableViewDataSource {
...
+            objectId: metadata?.fileId,
+            objectType: objectType,
+            previews: true) { (account, activities, errorCode, errorDescription) in
+                defer { disptachGroup.leave() }
+
+                guard errorCode == 0,
+                      account == self.appDelegate.account,
+                      let activity = activities.first,
+                      activity.idActivity > recentActivityId
+                else { return }
```
marinofaggiana, line 454: Always return at the start (DB empty) check it

marinofaggiana: Please before update the PR, do you the test.

thisIsTheFoxe: > Always return at the start (DB empty) check it

I'm not sure what you mean. If the DB is empty (aka `getLatestActivityId` == 0) I do return before requesting.. Feel free to elaborate..

marinofaggiana (approved):

PR #1746, opened 2021-11-10
Profile menu

thisIsTheFoxe:
Closes #1715 

Appearances:
- "Shared with you by [...]"
- Shared with table view (in Share view)
- Comments (in Activity view)
- Activities "Shared by [...]"

TODO: 
- [x] Implement for all avatar images
- [x] DCO
- [x] Test NC-talk integration

IMG_EB19372C1CBD-1

PR #1764, opened 2021-11-19
Profile improvements

thisIsTheFoxe:
Use `NCUtility.loadUserImage(for:)` wherever possible to load the locally generated placeholder instead when not connect to internet etc.
Fix bug, where the `sharedWithYouByView` was overlapping the search textfield.
Fix Hovercard SVG icon color

Related:
- #1746


iOSClient/Utility/NCUtility.swift
```diff
@@ -496,19 +496,21 @@ class NCUtility: NSObject {
         return  UIImage(named: "file")!.image(color: color, size: size)
     }
     
-    func loadUserImage(for user: String, displayName: String?, urlBase: String) -> UIImage {
-        var image: UIImage?
-        
-        let fileName = String(CCUtility.getUserUrlBase(user, urlBase: urlBase)) + "-" + user + "-original.png"
+    func loadUserImage(for user: String, displayName: String?, userUrlBase: String, original: Bool = false) -> UIImage {
+
+        //@marino: I'm not even sure the original is needed.. seems to work just fine with the 'normal' one
```
thisIsTheFoxe, line 501: @marinofaggiana can you confirm this?

marinofaggiana: I don't remember exactly this func (maked it several time ago) but seems works, yes.

Share/NCShareExtension.swift
```diff
@@ -263,7 +263,10 @@ class NCShareExtension: UIViewController, NCListCellDelegate, NCEmptyDataSetDele
 
         // PROFILE BUTTON
 
-        let image = NCUtility.shared.loadUserImage(for: activeAccount.user, displayName: activeAccount.displayName, urlBase: activeAccount.urlBase)
+        let image = NCUtility.shared.loadUserImage(
+            for: activeAccount.user,
+               displayName: activeAccount.displayName,
+               userUrlBase: String(CCUtility.getUserUrlBase(activeAccount.user, urlBase: activeAccount.urlBase)))
```
marinofaggiana, line 269: I don't like this very old func CCUtility.getUserUrlBase(...) I would replace it with a:

appDelegate.user + "-" + URL(string: appDelegate.urlBase).host

and remove the CCUtility.getUserUrlBase(...)  whats do you think ?

marinofaggiana:
The "original" is for the old CCManageAccount that required the "normal" image and not the (avatar)

Share/NCShareExtension.swift
thisIsTheFoxe: Agreed. I like to use computed vars for those things..

iOSClient/AppDelegate.swift
```diff
@@ -30,7 +30,7 @@ import Firebase
 import IHProgressHUD
 
 @UIApplicationMain
-class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate, TOPasscodeViewControllerDelegate, NCAccountRequestDelegate, NCViewCertificateDetailsDelegate {
+class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate, TOPasscodeViewControllerDelegate, NCAccountRequestDelegate, NCViewCertificateDetailsDelegate, NCUserBaseUrl {
```
marinofaggiana, line 34: NCUserBaseUrl ?

thisIsTheFoxe: Yea, I'm not happy with the naming as well.. if you have a better idea let me know..

iOSClient/Data/NCDatabase.swift
```diff
@@ -29,8 +29,12 @@ protocol DateCompareable {
     var dateKey: Date { get }
 }
 
-class tableAccount: Object {
+class tableAccount: Object, NCUserBaseUrl {
 
+//    @objc var userUrlBase: String {
+//        user + "-" + (URL(string: urlBase)?.host ?? "")
+//    }
+
```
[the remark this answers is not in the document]
thisIsTheFoxe: oh, lol

iOSClient/AppDelegate.swift
marinofaggiana: I don't undestand because you use a new class NCUserBaseUrl for a simple creation of name of avatar ...  user + "-" + (URL(string: urlBase)?.host ?? "")

thisIsTheFoxe: It's not a class it's a protocol. It just generalises how the user-url-id is created, for all types that can do so.

marinofaggiana: ja ja protocol (same) then we can, then, create a new file as NCCellProtocol.swift

marinofaggiana: of name .... NCUserFile -> var userFile: String , sounds good for you ?

thisIsTheFoxe: > ja ja protocol (same)

Actually, they're not and if you haven't seen it I highly recommend the "Crusty talk" to you..

About the name, I was more thinking of the protocol as a _capability_, but the more I think about it it's actually just means that the type has (or is part of) a user and a server url. So it makes sense as it is (to me). And I think that since it combines the user and url, they should be reflected in the name as well.

marinofaggiana: urlBase was the name of old (url) ownCloud driver objc of several year ago ... this project is many many years old and now seems simple but several year ago was a massacre... and many new technologies weren't there :)

marinofaggiana: a million things can be improved but let's take it easy ...

marinofaggiana (approved):

PR #1766, opened 2021-11-22
Add quick actions

### LINKED ISSUE #1765: Home Screen (long-press) Shortcuts
It would be nice to be able to long-press (or force-touch) the nextcloud app icon to access a shortcut menu. Most native apple apps use this. Some shortcuts that I think would be useful are:
- Scan Document
- Upload File/Photo/Video
- Favorites

The current shortcut menu looks like:
image

For reference, other apps look like:
image

thisIsTheFoxe:
- Close #1765 

Setup via code over static .plist to check if scanner is available (iOS 13+). Plus, an account is needed for it to work anyway.

### DIFF
--- a/iOSClient/AppDelegate.swift
+++ b/iOSClient/AppDelegate.swift
@@ -328,32 +328,32 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
 
         let shortcutItmes = UIApplication.shared.shortcutItems
-        if shortcutItmes == nil || shortcutItmes?.isEmpty == true {
-            let uploadFileItem = UIApplicationShortcutItem(
-                type: NCGlobal.QuickAction.uploadFile.rawValue,
-                localizedTitle: NSLocalizedString("_upload_file_", comment: ""),
-                localizedSubtitle: nil,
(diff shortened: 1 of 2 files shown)


marinofaggiana:
We use the shortcuts ? Apple did not want to delete them ?

thisIsTheFoxe:
@marinofaggiana I'm not sure what you mean, I didn't find any shoutouts in the app. Also, as far as I can see, `UIApplication.shared.shortcutItems` is not used anywhere..

marinofaggiana:
@thisIsTheFoxe no no, I mean: no body use the shortcuts ... :D I had read that Apple wanted to delete them from iOS

thisIsTheFoxe:
@marinofaggiana ok, but do you mean the *Shortcuts.app* or the technically called *Quick Actions* as mentioned in the issue? Cuz they are (besides the unlucky naming) not related afaik. 

And again, don't know where you read it, but as far as Apple goes I haven't heard anything about that officially. Actually, the fact that they are pushing Siri and Intents makes me think they have more plans with that.. especially, since the Shortcuts.app is now also available in Monterey.

marinofaggiana:
@thisIsTheFoxe yes was in the Apple plan remove 3D Touch and the quick actions because not used

PR #1790, opened 2021-12-01
Server trust

marinofaggiana:
Improved the check Trusted Challenge for don't verify the certificate saved when the certificate is Trusted


iOSClient/Networking/NCNetworking.swift
```diff
@@ -173,8 +173,10 @@ import Queuer
             
             var secresult = SecTrustResultType.invalid
             let status = SecTrustEvaluate(serverTrust, &secresult)
-            if errSecSuccess == status, let serverCertificate = SecTrustGetCertificateAtIndex(serverTrust, 0) {
-                    
+            let isServerTrusted = SecTrustEvaluateWithError(serverTrust, nil)
+
+            if let serverCertificate = SecTrustGetCertificateAtIndex(serverTrust, 0) {
```
thisIsTheFoxe, line 178: Can you combine the two `if`s? And then then use `serverCertificate` with `saveX509Certificate(serverCertificate, [...])` directly?

iOSClient/Networking/NCNetworking.swift
```diff
@@ -183,10 +185,14 @@ import Queuer
...
-                if let certificateSaved = NSData(contentsOfFile: certificateSavedPath), certificate.isEqual(to: certificateSaved as Data) {
+                if isServerTrusted {
                     return true
+                } else if status == errSecSuccess {
+                    // verify
+                    let certificateSavedPath = directoryCertificate + "/" + host + ".der"
+                    if let certificateSaved = NSData(contentsOfFile: certificateSavedPath), certificate.isEqual(to: certificateSaved as Data) {
+                        return true
+                    }
                 }
```
thisIsTheFoxe, line 196:
```suggestion
                    } // else: certificate not trusted, fallthrough
```
Ending with an `else-if` is a code smell. At the very least add a comment as to why there is no final `else`

thisIsTheFoxe (changes requested):

iOSClient/Networking/NCNetworking.swift
marinofaggiana: No finished yet, ...

iOSClient/AppDelegate.swift
```diff
@@ -535,62 +530,53 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-                    self.window?.rootViewController?.present(navigationController, animated: true)
-                }
-            }))
-            
-            window?.rootViewController?.present(alertController, animated: true, completion: {
-                self.timerErrorNetworking?.invalidate()
-            })
+        guard let currentHost = URL(string: self.urlBase)?.host else { return }
+        guard let pushNotificationServerProxyHost = URL(string: NCBrandOptions.shared.pushNotificationServerProxy)?.host else { return }
+        if host == pushNotificationServerProxyHost || host != currentHost { return }
```
thisIsTheFoxe, line 549: You could combine all of those into one `guard` statement

iOSClient/Networking/NCNetworking.swift
```diff
@@ -163,39 +161,49 @@ import Queuer
         let protectionSpace: URLProtectionSpace = challenge.protectionSpace
         let directoryCertificate = CCUtility.getDirectoryCerificates()!
         let host = challenge.protectionSpace.host
-        let pushNotificationServerProxyHost = URL(string: NCBrandOptions.shared.pushNotificationServerProxy)?.host
-            
+        let certificateSavedPath = directoryCertificate + "/" + host + ".der"
+        var isTrusted: Bool
+
+        defer {
+            #if !EXTENSION
```
thisIsTheFoxe, line 168: Use macros at the top most level possible. So, first `#if` then `defer`

iOSClient/Networking/NCNetworking.swift
```diff
@@ -163,39 +161,49 @@ import Queuer
...
+            #endif
+        }
+        
         print("SSL host: \(host)")
         
-        if let serverTrust: SecTrust = protectionSpace.serverTrust {
+        if let serverTrust: SecTrust = protectionSpace.serverTrust, let serverCertificate = SecTrustGetCertificateAtIndex(serverTrust, 0)  {
             
+            // extarct certificate txt
             saveX509Certificate(serverTrust, host: host, directoryCertificate: directoryCertificate)
```
thisIsTheFoxe, line 182: You can refactor this a bit to pass the certificate directly instead of the trust. Makes more sense semantically and you can remove the same check from the `saveX509Certificate` function (the one calulating `SecTrustGetCertificateAtIndex(trust, 0)`.

thisIsTheFoxe (changes requested):

iOSClient/Networking/NCNetworking.swift
marinofaggiana: oh! true great eye

thisIsTheFoxe (approved):
I can't validate that it works, bc I don't have a TLS server setup. But in general lgtm

PR #1839, opened 2021-12-28
Fix Share extension duplicates

thisIsTheFoxe:
1. move all files into to the app
2. check if file with same path exists
finally, ask user to resolve conflicts.
- Use `DispatchQueue` to keep track of all (un-)uploaded files. When done, dismiss.
- Don't let user rename with already existing filename, feedback welcome:
  -  just return and do nothing (current implementation of the PR)
  - show an error
  - disable rename button, if text matches one of the other files

TODO:
- [x] Test
- [ ] Error handling

### DIFF
--- a/Nextcloud.xcodeproj/project.pbxproj
+++ b/Nextcloud.xcodeproj/project.pbxproj
@@ -18,6 +18,11 @@
 		371B5A2E23D0B04500FAFAE9 /* NCMenu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 371B5A2D23D0B04500FAFAE9 /* NCMenu.swift */; };
 		3781B9B023DB2B7E006B4B1D /* AppDelegate+Menu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3781B9AF23DB2B7E006B4B1D /* AppDelegate+Menu.swift */; };
 		8491B1CD273BBA82001C8C5B /* UIViewController+Menu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8491B1CC273BBA82001C8C5B /* UIViewController+Menu.swift */; };
+		AF22B206277B4E4C00DAB0CC /* NCCreateFormUploadConflict.swift in Sources */ = {isa = PBXBuildFile; fileRef = F704B5E42430AA8000632F5F /* NCCreateFormUploadConflict.swift */; };
+		AF22B207277B4E4C00DAB0CC /* NCCreateFormUploadConflict.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = F704B5E22430AA6F00632F5F /* NCCreateFormUploadConflict.storyboard */; };
+		AF22B208277B4E4C00DAB0CC /* NCCreateFormUploadConflictCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = F704B5E82430C0B800632F5F /* NCCreateFormUploadConflictCell.swift */; };
+		AF22B209277B4E4C00DAB0CC /* NCCreateFormUploadConflictCell.xib in Resources */ = {isa = PBXBuildFile; fileRef = F704B5E62430C06700632F5F /* NCCreateFormUploadConflictCell.xib */; };
+		AF2AC8CC27620F9300F249C7 /* XLForm in Frameworks */ = {isa = PBXBuildFile; productRef = AF2AC8CB27620F9300F249C7 /* XLForm */; };
 		AF2D7C7C2742556F00ADF566 /* NCShareLinkCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = AF2D7C7B2742556F00ADF566 /* NCShareLinkCell.swift */; };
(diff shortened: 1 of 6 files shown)


marinofaggiana:
@thisIsTheFoxe you are completely broken the code ... how I can check it out ? it was necessary to have to rewrite it ?

thisIsTheFoxe:
@marinofaggiana 
1. done (see discussion in the 2nd issue)
2. done, the error is still needed when renaming a file since you can't have the same filename twice
4. I've tested it with a bunch of files and seems to work fine

marinofaggiana:
> @marinofaggiana
> 
> 1. done (see discussion in the 2nd issue)

I don't look the "..." for 1 file ?

Schermata 2022-01-08 alle 11 21 21
Schermata 2022-01-08 alle 11 21 06

> 3. I've tested it with a bunch of files and seems to work fine

Good !

- new issue, counter error: Upload file 0/1 ?

thisIsTheFoxe:
> I don't look the symbol rename for 1 file ?

The idea was that a person cannot remove the last file, since then there'd be nothing left to upload. But to not overcomplicate things I reverted it as it was before. 

> new issue, counter error: Upload file 0/1 ?

Seemed fine to me, 0 / 1 file is uploaded, so it's uploading the first file. But I see how that could be confusing.. fixed it.

marinofaggiana:
> > I don't look the symbol rename for 1 file ?
> 
> The idea was that a person cannot remove the last file, since then there'd be nothing left to upload. But to not overcomplicate things I reverted it as it was before.

Eventually we can consider replacing (for only 1) the "..." with the "pencil" to change only the name? what do you think ?

thisIsTheFoxe:
Not quite sure what you mean by "eventually".. why would we postpone it?

But, I also considered that. And that was why I said _not to overcomplicate things_. It's an idea but I don't really think it's worth it..

marinofaggiana:
Ok, can you rebase this PR so I can test it with my HW ?

thisIsTheFoxe:
Done.

marinofaggiana:
@thisIsTheFoxe I have try with my iPhone 12 Pro Max iOS 15.2, two image, one of 5 MB and one of 500 kb, when detect that the file exists crashed ...

Thread 1: EXC_RESOURCE RESOURCE_TYPE_MEMORY (limit=120 MB, unused=0x0)

as I imagined ...

thisIsTheFoxe:
Can confirm 👍 
I tried it with like 20+ images but I had none that was bigger than 5MB :D

Anyway, downsampling them before displaying solves it. Tried it with a 8k (17MB) with a max memory usage of like 20MB.

marinofaggiana:
Good, seems now works but still some problems (I don't know if it was before but it doesn't matter).
Upload several file with different size, example 10 file from 5 kb until video over 100 MB (chunk) ...  the IHProgressHUD seems not to work well and if I cancel the share during the upload the files remain in queue ( try and open the app ) I don't think it's correct ... is better use the Upload without the background (??? this is dangerous),  what's do you think ? 

if you don't know how to do it, I'll gladly help you.

marinofaggiana:
Note. I checked the code .. in fact it didn't, it was simpler but more functional, better to send the files one by one; as I told you before making changes ask me, many things always have a certain logic behind them. please ask! 
Doing so yes but half a step forward and 4 back.

marinofaggiana:
+ Added bug: Screen capture of iPhone and send to Share ... missing the extension.

thisIsTheFoxe:
Hmm.. for the record, I still think there's a big benefit in parallelizing here, especially if you have many smaller files.. it takes very long..
But I also see how not being able to cancel request is quite a drawback.. I'll revert..

> Screen capture of iPhone and send to Share ... missing the extension

Not sure what you mean. Screenshots as well as screen recordings (saved as video in Photos) can be shared without problems for me.

marinofaggiana:
Good, now works but remain the upload in progress, if the user close the share the upload remains ... and it is the case to eliminate it .... so you can detect when the extension is closed and remove the metadata  before the exit.

line 358 let metadata = uploadMetadata.removeFirst()

thisIsTheFoxe:
@marinofaggiana I'm not sure what you mean. I use `uploadStarted` to ensure that the extension is active. If the user cancels it, no further uploads are started..

marinofaggiana:
The problem is the following:

The user selects File n where n can only be one, he realizes that the operation is wrong (he is sending the file to an incorrect location or the file is wrong, for example it is huge) by forcibly closing the share all files must be deleted and the uploaded file must be stopped and deleted.
Now the upload file remains in the queue and when you open the Nextcloud application it resumes.

I don't know if I'm clear now, try it yourself, the important thing is that no file remains pending, if I close everything it must be deleted.

marinofaggiana:
@thisIsTheFoxe do you have done some experiment to see if it is possible to intercept the dismiss of UIViewController without the cancel button (Swipe down the view)? (but ... I don't think it can be)

marinofaggiana (commented):
Seems works.

marinofaggiana (approved):

PR #1852, opened 2022-01-16
fix: folder creation space at start/end

McP4nth3r:
This fix is a adaption of this issue https://github.com/nextcloud/server/issues/5843, sadly it was still possible to create a folder with whitespaces at start/end to break the windows sync.


thisIsTheFoxe (changes requested):
Hey,
Thanks a lot for your contribution!

The idea is good. 
However, I noticed that the [linked issue](https://github.com/nextcloud/server/issues/5843#issuecomment-460043910) mentioned  that the same behaviour should be expected for folder as well as for files, right?
And please also note that it is still possible to rename folders (and files) to have leading or trailing spaces.

And finally (just for the record), while sanitising the filename at upload time is very safe, it could _arguably_ be a better UX to not even allow the user to enter bad file/folder names in the first place. But messing with user input can also cause confusion and the web UI doesn't do it either.. so I think it's fine, just an idea.. ^^

McP4nth3r:
I will take a look at rename files/folders in the nex days :) 

i'm totally new to Swift language, and will try to implement this 👌🏻

a comment to your last point: 
the web-ui doesn't disallow whitespaces when i enter whitespaces than the name changed to trimmed name

thisIsTheFoxe:
Sounds good 👍
And no worries.. if you need help or have any questions feel free to ping me :)

thisIsTheFoxe (changes requested):
Hi,
This looks very promising.. 👍
For creating new files maybe you could have a look at:
https://github.com/nextcloud/ios/blob/d33196a70f4cfe495260d7c0f56a998560f3815d/iOSClient/Data/NCManageDatabse%2BMetadata.swift#L140

Also, I noticed you selected the default "master" branch as base branch. We usually only work on "develop" and keep the master branch stable. To fix the commit history you could run:
```
git pull origin develop
```
..and fix the conflicts. But for smaller changes like that it might be easier to just create a new branch from "develop" and apply your changes there and then force push it to this remote branch (or push it as a new branch and make a new PR).

McP4nth3r:
Ok I will change the branch in the evening👌🏻. In createMetadata the fileName is the fileName on Server and the fileNameView is only the display name? So I could change the raw fileName parameter to have all the same? 
Is there a Plan when the new release with the fix if all is working is applying to the AppStore? 😉

McP4nth3r:
open new one with develop branch 👌

PR #1856, opened 2022-01-19
NCTrash select

### LINKED ISSUE #1782: Trash - Option „Select“ or „Select all“ in order to restore or delete multiple files is missing
Frequency
Every Time

Actions Performed
Prerequisite: A large amount of files have already been added to the trash folder.
1. Open the app.
2. Login with a valid account.
3. Tap on the „More“.
4. Tap on „Trash“.
5. Try to restore multiple files at the same time. Tap on the threes dots icon in the row of the sorting options.
6. Check the available functions.

Expected Result
Like in every other folder list view in this app there exists a possibility to select files or select all at the same time within the three dots menu. It is easy to restore multiple files simultaneously.

Actual Result
The option „Select all“ or simply „Select“ in order to restore multiple files at the same time is missing in the three dots menu. This means that the user has to restore every file one by one.

thisIsTheFoxe:
Fix #1782 

- Make NCTrash subclass of `NCCollectionViewCommon`
- add options to restore / delete for all and selected files


marinofaggiana:
@thisIsTheFoxe can you rebase this PR ?

marinofaggiana:
Test error:

- Select two or more files
- Menu: Delete selected files
- Removed only one file and after
- 500: Error ...

marinofaggiana:
Updated now to Nextcloud 23.0.1 RC3 and works as expected, good I continue with test

marinofaggiana (approved):

Issue #1733, opened 2021-10-31
Easier renaming in upload via Share Menu


thgoebel:
### Steps to reproduce

1. Click "Share" in another app and select Nextcloud in the Share Menu
2. Try to rename the file before upload

### Expected behaviour

A simple click on the filename should bring up the popup with the TextEdit field (1 step).

### Actual behaviour

I need to click the ... and then "Rename file" (2 steps).

### Screenshots

<img width="50%" src="https://user-images.githubusercontent.com/33295590/139577640-d943b94a-fab5-41f9-a2e1-e5b5bca960f8.png">

### Reasoning or why should it be changed/implemented?

1. The action in uncessarily hidden away.
2. Clicking on the file name to change it is natural.
3. Why is the ... menu even needed? In what situation would I need to *delete* a file that I am *uploading*? Where would it even be deleted from? The 3rd app that I am sharing it from?? The Nextcloud server - where it most likely does not exist yet??

### Environment data

**iOS version:** iPadOS 15,1

**Nextcloud iOS app version:** 4.0.6.0

thisIsTheFoxe:
Hi,
Thank you for your feedback. And I can see how we might be able to simplify the renaming process.

About your 3rd point, *"Remove file"* would probably fit a lot better here. Because it is possible to share multiple files at once. E.g. in the iOS Files.app -> "..." -> Select -> Share. In that case you can remove some of the files you previously selected and still share the rest of them without going back. When sharing a single file, it is somewhat unnecessary since it does the same as "Cancel".

marinofaggiana:
I confirm the delete file is a very good option which does not bother. I think we can close this issue

thgoebel:
I absolutely agree with you that the remove action makes sense!
Still, the fact that I didn't get it means that it is probably worth renaming "Delete file" -> "Remove file" to make more clear what it does :). Note that making sure that all translations are renamed might be tedious. :grimacing: 

Some other ideas for the UX (I would love the renaming to be more easily accessible!):

Idea 1:
- Rename: by clicking on the file name.
- Remove: by clicking on a dedicated icon (e.g. Ⓧ or 🗑️).
- Pros: fastest. 1 click for both actions.
- Cons: Icons-only, no explicit text. And for my not-really-technologically-versed parents it might be not intuitive to click on the file name to change it.
[...]

PR #1857, opened 2022-01-20
fix share extension

marinofaggiana:
Signed-off-by: marinofaggiana <ios@nextcloud.com>

Read comments.


Share/NCShareExtension.swift
```diff
@@ -177,17 +177,22 @@ class NCShareExtension: UIViewController {
         collectionView.reloadData()
         tableView.reloadData()
     }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        self.cancel(with: .cancel)
+    }
```
marinofaggiana, line 185: Please test it: if the user dismiss the share without the cancel button (to drag down) the current upload metadata is correctly removed

Share/NCShareExtension.swift
```diff
@@ -177,17 +177,22 @@ class NCShareExtension: UIViewController {
...
-        let filePath = CCUtility.getDirectoryProviderStorageOcId(metadata.ocId, fileNameView: metadata.fileNameView)!
-
-        NCManageDatabase.shared.deleteMetadata(predicate: NSPredicate(format: "ocId == %@", metadata.ocId))
-        NCNetworking.shared.uploadRequest[filePath]?.tasks.forEach({ $0.cancel() })
+        for metadata in uploadMetadata {
+            let filePath = CCUtility.getDirectoryProviderStorageOcId(metadata.ocId, fileNameView: metadata.fileNameView)!
+            NCManageDatabase.shared.deleteMetadata(predicate: NSPredicate(format: "ocId == %@", metadata.ocId))
+            NCNetworking.shared.uploadRequest[filePath]?.tasks.forEach({ $0.cancel() })
+        }
         extensionContext?.cancelRequest(withError: error)
```
marinofaggiana, line 196: if the user has remove the last or one file and close the share the app crashed without call the extensionContext?.cancelRequest

Share/NCShareExtension.swift
thisIsTheFoxe: I did have some issues with calling cancel multiple times before, even causing runtime crashes..
But I think the main problem is this wired behaviour when trying to dismiss:

ScreenRecording

Share/NCShareExtension.swift
thisIsTheFoxe: I mean the user cannot remove metadata while it's being uploaded but yea.. makes sense 👍

Share/NCShareExtension.swift
```diff
@@ -370,6 +375,9 @@ extension NCShareExtension {
         // CHUNCK
         metadata.chunk = chunckSize != 0 && metadata.size > chunckSize
 
+        let status = NSLocalizedString("_upload_file_", comment: "") + " \(counterUploaded + 1) " + NSLocalizedString("_of_", comment: "") + " \(filesName.count)"
+        IHProgressHUD.set(status: status)
+        
         NCNetworking.shared.upload(metadata: metadata) { } completion: { errorCode, _ in
```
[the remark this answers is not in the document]
thisIsTheFoxe: Hmm.. I thought about that too.. but this will also reset the progress, right? And it causes the progress to glitch.. Maybe we can call it even before that..

ScreenRecording2

thisIsTheFoxe (changes requested):
Good ideas, but it's not quite there yet..
Also, please remember the listing issues, thx :)

marinofaggiana:
@thisIsTheFoxe probably find a good compromise. please check

marinofaggiana:
I cant' look the glitching because I don'u use a server in the same local network, but with this last push I have fixed, test it.

thisIsTheFoxe (changes requested):
Looks like the progress glitch is fixed.

- Fix the lint warnings
- With a slow network the HUD doesn't show:
https://user-images.githubusercontent.com/18512366/150777799-3f937e5b-613d-489c-b852-dbf2e9eb5bcc.mov

marinofaggiana:
@thisIsTheFoxe strange ... from me do not happen... can you make two test ?

- without the view "replace", with new files
- with replace but added a DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) before the Upload call

Thanks

marinofaggiana:
Note: why when you touch on Cell (not "...") appear the rename ?

Share/NCShareExtension.swift
```diff
@@ -369,7 +375,12 @@ extension NCShareExtension {
 
         // CHUNCK
         metadata.chunk = chunckSize != 0 && metadata.size > chunckSize
-
+        
+        if hud {
```
thisIsTheFoxe, line 379: Do you need this variable? Can't you just test for `counterUploaded == 0`?

thisIsTheFoxe (changes requested):
Thanks. The HUD glitch is fixed.

> Note: why when you touch on Cell (not "...") appear the rename ?

Please check out the issue linked in my original PR: #1733

I also narrowed the bug down to (happens w/ and w/o conflict):
1. _Open In_ > _Nextcloud_
2. _Upload_ (HUD is shown)
3. _Cancel_
4. Again select _Nextcloud_
5. _Upload_ (HUD isn't shown)

Also, please remember that we have a Lint CI now :)

marinofaggiana:
@thisIsTheFoxe I have update the IHProgressHUD, this is not definitive but it works, when I have time I will want to make major changes.

Share/NCShareExtension.swift
```diff
@@ -137,9 +137,8 @@ class NCShareExtension: UIViewController {
         }
 
         // HUD
-        IHProgressHUD.set(viewForExtension: self.collectionView)
+        IHProgressHUD.set(viewForExtension: self.view)
```
thisIsTheFoxe, line 140: This seems to fix the bug, but I don't like this solution because:
If you have multiple files that take a while to upload you want to be able to scroll through the bottom file-list to check if everything's there / how many files are left etc. Showing the HUD over the entire view makes the list non-intractable and it isn't obvious why and thus  feels like a bug.

thisIsTheFoxe (changes requested):
I think there is a better solution to this..

Share/NCShareExtension.swift
marinofaggiana: - The fix (patch) in on IHProgressHUD.
- If you use self.collectionView and move the collectionView the HUD follow the scroll and the effect it's horrible
- The touch on Cell works during upload (every button must be disabled), look:



The self.view works perfectly and the number on HUD show at what number is the upload (and is a good info), nobody would scroll during a file upload.

If the bug is solved, merged.

thisIsTheFoxe: > - The fix (patch) in on IHProgressHUD.
> - If you use self.collectionView and move the collectionView the HUD follow the scroll and the effect it's horrible
> - The touch on Cell works during upload (every button must be disabled), look:

@marinofaggiana 
- I don't understand what you mean
- That's not possible because the HUD view covers the collection view and interrupts any user input. If you move the collection view the entire modal VC moves with it. It's the same behaviour when using the entire view
- I assume that's the create folder? Correct, that would need to be disabled, or even better remove the directory cell entirely when uploading

marinofaggiana: Look:

Simulator Screen Shot - iPhone 13 - 2022-01-26 at 11 38 15

thisIsTheFoxe (approved):
I still think it's odd that the HUD blocks the interaction with the bottom file list. Maybe for the future we could think about adding a mask / background to the HUD to make it clear that the view is blocked.

PR #1862, opened 2022-01-24
fix: folder creation with whitespaces at start/end

McP4nth3r:
This fix is a adaption of this issue nextcloud/server#5843, sadly it was still possible to create a folder with whitespaces at start/end to break the windows sync.

requested new because wrong target branch https://github.com/nextcloud/ios/pull/1852

Signed-off-by: McP4nth3r <burg41@yahoo.de>

### DIFF
--- a/iOSClient/Data/NCManageDatabse+Metadata.swift
+++ b/iOSClient/Data/NCManageDatabse+Metadata.swift
@@ -141,6 +141,8 @@ extension NCManageDatabase {
 
         let metadata = tableMetadata()
         let resultInternalType = NCCommunicationCommon.shared.getInternalType(fileName: fileName, mimeType: contentType, directory: false)
+        
+        let fileName = fileName.trimmingCharacters(in: .whitespacesAndNewlines)
 
         metadata.account = account
         metadata.chunk = false
--- a/iOSClient/Main/Create cloud/NCCreateFormUploadDocuments.swift
+++ b/iOSClient/Main/Create cloud/NCCreateFormUploadDocuments.swift
@@ -275,7 +275,7 @@ import XLForm
         } else {
 
             //Trim whitespaces after checks above
-            fileNameForm = fileNameForm.trimmingCharacters(in: .whitespacesAndNewlines)
+            fileNameForm = (fileNameForm as! String).trimmingCharacters(in: .whitespacesAndNewlines)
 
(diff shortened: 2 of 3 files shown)


thisIsTheFoxe (changes requested):
> So I could change the raw fileName parameter to have all the same?
Is there a Plan when the new release with the fix if all is working is applying to the AppStore? 😉

I'm not sure why `fileNameView` is unused here.. but, I guess it's fine for now and we'll fix that another time..
We'll try our best to get the new version out there before February but no promises.. 😄

Good job so far. And I'm sorry I missed it before, but we create (text, office) documents a little differently. Could you please also add a check to:
https://github.com/nextcloud/ios/blob/650690cabb10fe70217649654eeef799de807a0d/iOSClient/Utility/CCUtility.m#L1277

McP4nth3r:
I would like to implement this, but i don't understand what you mean with add a check to.
So did you mean that I should add a check for whitespaces or convert string with withespaces to string string without?

thisIsTheFoxe:
@McP4nth3r I did mean to _check_ if the string contains whitespaces (which is done automatically by the trim function)  and convert it just as you did before.. :)
However, I am sorry once again.. it seems like that Obj-C function I mentioned is used a bit too much for such a change and would probably create some other issues.
Now, the variables that are important for creating new files are `fileNameForm` and `fileName` as you can see below. Could you make sure those are trimmed as well?

Thanks again for your help, it's really appreciated! ^-^

https://github.com/nextcloud/ios/blob/d33196a70f4cfe495260d7c0f56a998560f3815d/iOSClient/Main/Create%20cloud/NCCreateFormUploadDocuments.swift#L297
https://github.com/nextcloud/ios/blob/d33196a70f4cfe495260d7c0f56a998560f3815d/iOSClient/Main/Create%20cloud/NCCreateFormUploadDocuments.swift#L313

thisIsTheFoxe (changes requested):
Thanks!
The last thing is that `fileNameForm` is of type `Any`. In general it would be better to use safe casting in the `guard` statement and refactor the rest.
But to not overcomplicate things here, we can just use a force cast for now so the build success.. :)

McP4nth3r:
Thanks for the tip :)

thisIsTheFoxe (approved):
Looks good to me.. 👍

PR #1908, opened 2022-03-01
remove autoreleasepool in image colorized 

marinofaggiana:
- some crash registered in the autoreleasepool invoking, useless in this old func


thisIsTheFoxe (approved):
lgtm 👍

PR #1919, opened 2022-03-08
Fix upload files of size 0

### LINKED ISSUE #803: Files with size 0 behave weirdly
I've noticed that the app behaves weirdly when dealing with files that have no data (fileSize 0). I've traced this to this check which is used all over the app. Notice how it checks fileSize > 0. Is this intended? Wouldn't be checking fileSystemNumber or something yield better results?

CCUtility.m:926:
```
+ (BOOL)fileProviderStorageExists:(NSString *)fileID fileNameView:(NSString *)fileNameView
{
    NSString *fileNamePath = [self getDirectoryProviderStorageFileID:fileID fileNameView:fileNameView];
    
    unsigned long long fileSize = [[[NSFileManager defaultManager] attributesOfItemAtPath:fileNamePath error:nil] fileSize];
    
    if (fileSize > 0) return true;
    else return false;
}
```

### LINKED ISSUE #1882: Can't upload empty files
### Steps to reproduce

0. create a empty file (0 byte) e.g. by using the app: *+ -> Create text document* and save it in e.g. the Files.app 
1a) + -> Upload file -> select the empty file
1b) Go to the file, Open In -> Nextcloud -> Upload

### Expected behaviour
The empty file is being uploaded and will be displayed as such on Nextcloud

### Actual behaviour
In 1.1 an error appears with the key *err_file_not_found*.
In 1.2 the app gets stuck uploading. Furthermore, if multiple file are uploaded the app gets stuck and also doesn't upload the other files.

Nextcloud allows files with 0 byte to be created (mentioned in point 0) and the web UI also allows uploading empty (0 byte) files as well.

### Environment data

**iOS version:** e.g. iOS 15

**Nextcloud iOS app version:** 4.2.2.1

**Nextcloud version:** 23.0.1

### LINKED ISSUE #1923: Document scan fails after last app update
Scanning a document fails after scanning the image, adding to the scan, and hitting save.  Failure occurs right when it asks you to save or delete the image.

image

thisIsTheFoxe:
Fix #1882 
Fix #803 
Fix #1923 

- **+** > Upload File > Select file with 0 bytes, didn't work before
- Fix bugs where the metadata size isn't set correctly (Scan documents & Voice memo)

### DIFF
--- a/iOSClient/AppDelegate.swift
+++ b/iOSClient/AppDelegate.swift
@@ -873,13 +873,13 @@ extension AppDelegate: NCAudioRecorderViewControllerDelegate {
 
     func didFinishRecording(_ viewController: NCAudioRecorderViewController, fileName: String) {
 
-        guard let navigationController = UIStoryboard(name: "NCCreateFormUploadVoiceNote", bundle: nil).instantiateInitialViewController() else { return }
-        navigationController.modalPresentationStyle = UIModalPresentationStyle.formSheet
-        let appDelegate = UIApplication.shared.delegate as! AppDelegate
-
-        let viewController = (navigationController as! UINavigationController).topViewController as! NCCreateFormUploadVoiceNote
-        viewController.setup(serverUrl: appDelegate.activeServerUrl, fileNamePath: NSTemporaryDirectory() + fileName, fileName: fileName)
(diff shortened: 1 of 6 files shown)


marinofaggiana:
Seems works, required tet on TestFlight.

marinofaggiana (approved):

PR #1926, opened 2022-03-14
Fix QuickLook overriding files

### LINKED ISSUE #1629: Overwriting original in annotation mode does not work
### Steps to reproduce

1. Use Nextcloud Liquid for iOS 4.0.1.1
2. Open context menu of pdf file (three dots)
3. Select annotate(?)/change with the pencil icon in the circle (DE: Ändern)
4. Draw something in the document
5. Press Done and select overwrite original
6. Reopen the document to check if changes were saved
7. Be sad you lost 2h of annotations
(Saving a copy works)
### Expected behaviour

When overwriting the original, it should be writting to storage and nextcloud. Changes shouldn't disappear. 

### Actual behaviour

Nothing is saved, all is lost. 

It means data loss, which is really frustrating and damaging to the perception of NC quality.

### Environment data

**iOS version:** e.g. iOS 14.6

**Nextcloud iOS app version:** Nextcloud Liquid for iOS 4.0.1.1

Operating system: Linux 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64

Webserver: nginx/1.18.0 (fpm-fcgi) + nginx reverse proxy

Database: mysql 10.3.29

PHP version: 7.4.3

Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, session, standard, sodium, cgi-fcgi, mysqlnd, PDO, xml, apcu, bcmath, calendar, ctype, curl, dom, mbstring, FFI, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, intl, json, exif, mysqli, pdo_mysql, apc, posix, readline, redis, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, xmlreader, xmlwriter, xsl, zip, Phar, Zend OPcache

Nextcloud version: 21.0.2 - 21.0.2.1

thisIsTheFoxe:
Fix #1629 

[Flow Diagram](https://user-images.githubusercontent.com/18512366/158338242-280fffdb-4588-4733-92ba-c82d500bdce9.png)

- ask for saving option after view dismisses in order to catch all last changes
- remove cancel option, because view is already dismissed, can't cancel anymore
  - Alternative: Add option "Keep Editing" to re-open the temporary copy
 - remove auto-selecting the edit button for images
   - it's too hacky and doesn't work when the VC is not customly hosted in a NavVC. So either have the auto-select or have saving files work correctly. Imo it's better to have a slightly more complex UX than to have users loose their work or edits they made.

### Debugging / Testing Note
As mentioned in the issue, there seems to be a bug with the Xcode iOS Simulator where the file at `modifiedContentsURL` doesn't exist or was deleted.

### DIFF
--- a/iOSClient/Viewer/NCViewerMedia/NCViewerMediaPage.swift
+++ b/iOSClient/Viewer/NCViewerMedia/NCViewerMediaPage.swift
@@ -104,9 +104,8 @@ class NCViewerMediaPage: UIViewController {
         NotificationCenter.default.addObserver(self, selector: #selector(applicationDidBecomeActive(_:)), name: NSNotification.Name(rawValue: NCGlobal.shared.notificationCenterApplicationDidBecomeActive), object: nil)
     }
 
-    override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-
+    deinit {
+        print("#deinit NCViewerMediaPage")
         // Clear
         if let ncplayer = currentViewController.ncplayer, ncplayer.isPlay() {
             ncplayer.playerPause()
@@ -117,11 +116,11 @@ class NCViewerMediaPage: UIViewController {
 
         metadatas.removeAll()
         ncplayerLivePhoto = nil
-        
+
         #if MFFFLIB
         MFFF.shared.dismissMessage()
         #endif
-        
+
         // Remove Observer
         NotificationCenter.default.removeObserver(self, name: NSNotification.Name(rawValue: NCGlobal.shared.notificationCenterDeleteFile), object: nil)
         NotificationCenter.default.removeObserver(self, name: NSNotification.Name(rawValue: NCGlobal.shared.notificationCenterRenameFile), object: nil)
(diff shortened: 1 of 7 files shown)


marinofaggiana:
- Open an image
- Modify
- save Image
- the image do not update and if I try to swipe to next or previous image crashed
- Crash line 485 in NCViewerMediaPage

marinofaggiana (changes requested):
Crash line 485 NCViewerMediaPage

iOSClient/Viewer/NCViewerMedia/NCViewerMediaPage.swift
```diff
@@ -117,11 +116,11 @@ class NCViewerMediaPage: UIViewController {
 
         metadatas.removeAll()
```
thisIsTheFoxe, line 117: @marinofaggiana fixed it and also a potential memory leak..
 I'm not sure why you clear the array here. After the view disappears / is reinitialised it shouldn't matter, should it?
Also, I don't think we need to remove all the observers that use selectors. wdyt?

marinofaggiana: @thisIsTheFoxe  theoretically not, but many recommend it ... so it doesn't hurt, but the refresh after the modify still do not works. Do you have try it ?

marinofaggiana (changes requested):
Refresh after the modify still do not works.

marinofaggiana:
> @marinofaggiana I noticed that too. Please note that this also doesn't work in the current version of the app (tried with 4.4.0.2). Also, note that to fix that we would also need to reload the entire PDF view as well.. I suggest in order to move along we make another issue and PR for that..

OK, Don't worry I create a new PR (after) for improved the "refresh", so please you don't work in the Viewer component because I have several modify in list & TO-DO.

marinofaggiana (approved):

PR #1933, opened 2022-03-22
Update the viewer after modify

marinofaggiana:
- after modify image reload the viewer

### DIFF
--- a/iOSClient/Viewer/NCViewerMedia/NCViewerMedia.swift
+++ b/iOSClient/Viewer/NCViewerMedia/NCViewerMedia.swift
@@ -194,12 +194,15 @@ class NCViewerMedia: UIViewController {
     // MARK: - Image
 
     func reloadImage() {
-        loadImage(metadata: metadata) { _, image in
-            self.image = image
-            // do not update if is present the videoLayer
-            let numSublayers = self.imageVideoContainer.layer.sublayers?.count
-            if numSublayers == nil {
-                self.imageVideoContainer.image = image
+        if let metadata = NCManageDatabase.shared.getMetadataFromOcId(metadata.ocId) {
+            self.metadata = metadata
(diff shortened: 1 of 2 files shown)


iOSClient/Viewer/NCViewerMedia/NCViewerMedia.swift
```diff
@@ -196,6 +193,20 @@ class NCViewerMedia: UIViewController {
 
     // MARK: - Image
 
+    func reloadImage() {
+        if let metadata = NCManageDatabase.shared.getMetadataFromOcId(metadata.ocId) {
```
thisIsTheFoxe, line 197: Please use `guard` for early returns

iOSClient/Viewer/NCViewerMedia/NCViewerMediaPage.swift
```diff
@@ -228,6 +233,22 @@ class NCViewerMediaPage: UIViewController {
         }
     }
 
+    @objc func uploadedFile(_ notification: NSNotification) {
+
+        if let userInfo = notification.userInfo as NSDictionary? {
+            if let ocId = userInfo["ocId"] as? String, let metadata = NCManageDatabase.shared.getMetadataFromOcId(ocId), let errorCode = userInfo["errorCode"] as? Int {
+                if errorCode == 0, let index = metadatas.firstIndex(where: {$0.ocId == metadata.ocId}) {
```
thisIsTheFoxe, line 240: Again, please use _one_ `guard` instead of cascading `if`s :)
Makes it clearer and easier to read..

thisIsTheFoxe (changes requested):
Neat!
Also looks like I was wrong about the PDF quicklook.. works already :)

thisIsTheFoxe (approved):
#wontfix for now because timing

Issue #1814, opened 2021-12-11
Raw photos are uploaded as JPG when "most compatible" option is on


mhoeher:
### Steps to reproduce

1. Enable auto-upload, using the "most compatible" option in the advanced settings.
2. Take some raw photos.
3. Let the app upload them to the NextCloud server.

### Expected behaviour

The raw photos should be uploaded as-is. I think this is what most users expect: You use the default camera app for your "usual photo taking" and switch to another one (or enable ProRAW) when you want to take some "serious" ones you expect to post process later on.

Alternatively, if - for whatever reason - it makes sense to convert these images, then they should be converted to proper JPG files. However...

### Actual behaviour

The images are uploaded as JPG files, however, they get the `*.dng` file extension instead. This later on leads to issues when trying to open these images, as some programs use the file name extension and expect a raw photo inside - which isn't the case.

### Reasoning or why should it be changed/implemented?

The current behaviour generates "broken" files - they cannot be opened by some apps (unless you start renaming them to `*.jpg`).

As described further up, my expectation was that raw photos are not converted, as this makes using raw images in the first place useless.

### Environment data

**iOS version:**  iOS 15.1.1

**Nextcloud iOS app version:** 4.1.0.17

**Server operating system:** Fedora Linux

**Web server:** Apache (based on the docker-compose example by NextCloud).

**Nextcloud version:** 22.1.1

shawnhwei:
I think this happens regardless of whether you use the "most compatible" setting. I use the "high efficiency" setting and observe the same behavior (regular JPEG with DNG extension uploaded).

danfate:
Even if  "most compatible" option is turned off, the issue still exists.

mhoeher:
Hi @danfate,

you are right. I just checked my recently uploaded images and all of them - despite some of them are DNG on the phone - have been uploaded as JPG (with the "Most Compatible" option turned on).

Issue #833, opened 2019-04-10
Support RAW image upload


wyager:
### Expected behaviour                     

Importing a RAW or RAW+JPG image from the camera roll should import the whole RAW image                                                                                                                                                                                                                                                                                                                                                                                                                                                      

### Actual behaviour           

The app only imports the JPG part  (full size if available, preview size if not)                                                                                                                                                                                                                                                                                                                                                                                                                                                                   

### Steps to reproduce             

Import a RAW image from a camera and upload it                                                                                                                                                                                                                                                                                                                                                                                                                                                                

### iOS version

12.1.4

### App version

2.23.3.4

### Server configuration

15.0.5, installed via snap

**Operating system**:

Ubuntu 18.04



I know it's possible to import the whole image, including RAW data, since some apps like affinity photo can do it. There's probably just some flag you have to pass or something to prevent iOS from handing you JPG data only.

richardtsai:
A relevant but not identical issue: Auto-Upload seems to convert the DNG raw images to JPEGs but still keep the `.dng` extensions, which makes previewing fail.
![image](https://user-images.githubusercontent.com/1297550/108618073-0bdd6400-7456-11eb-8c0f-6b6426f1e267.png)

PR #1940, opened 2022-03-30
proraw photo uploading support

zcw159357:
Signed-off-by: zcw159357 <zcw159357@gmail.com>

EDIT:
- Ignore edits made in Photos.app for any DNG images!
- Alternative:
  - Only ignore edits if `getFormatCompatibility` is disabled

### DIFF
--- a/iOSClient/Utility/CCUtility.m
+++ b/iOSClient/Utility/CCUtility.m
@@ -1392,13 +1392,19 @@
                 }
             };
             
+            NSString *extensionAsset = [[[asset valueForKey:@"filename"] pathExtension] uppercaseString];
+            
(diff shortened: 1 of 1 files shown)


zcw159357:
#1814 #833

zcw159357:
Hi @thisIsTheFoxe ,  I tried using PHImageRequestOptionsVersionOriginal and it works nicely, sorry I'm not a ios developer and I'm not familiar with the apis. I see fileName was accessed via valueForKey so I just go for it....

zcw159357:
Yes, saving both the raw file and the edit data may be a better way, but that'll require a lot more changes...

PR #1968, opened 2022-04-25
Files lock 🔒

### LINKED ISSUE #1888: Implement File Locking
- https://github.com/nextcloud/server/issues/21914
- https://github.com/nextcloud/files-clients/issues/2

### **If** file locking is enabled / available in the server capabilities:
- Add File (Un-)Lock action to context menu & NCMenu.
- Add file lock indicator for locked files
- If file is locked, only allow read-only actions
- If the user tries to open a locked file, show an alert and open it read-only

thisIsTheFoxe:
Requires:
- https://github.com/nextcloud/ios-communication-library/pull/113

Close #1888

---

### General
- [x] Display time left instead of date+time: "30 minutes left"
- [x] "Locked by Text" → "Locked by Text app"
	- [x] Nice to have: Add app icon if it’s locked by an app (if available in the project)
- [x] Action menu placement
	- [x] "Lock file" action should go below "Details" in action menu
	- [x] Lock info part all the way to the top
	- [x] "Unlock file" directly below the info part
- [x] Detect error code when you "Unlock file" and is already unlocked, don't show error

### iOS
- [x] Save dialog of "Modify" flow: Add some explanation text: "Unable to save because the original file is locked"
---

### Additional thoughts
- Favourites don't work for locked files (nextcloud/files_lock#68)
  - a) leave as is and let server return an error
  - b) remove favourite option for locked files
- "*Text app*" vs "*Text.app*"
- Maybe show lock status in detail (share) view somehow?
- ~~**Currently**, the etag doesn't seem to change when the lock runs out,~~ should additionally check on client side or rely only on etag?
- Only one client side lock error. Potentially use detailed error descriptions from server?


IdkSaveme (approved):

marinofaggiana (changes requested):
UIX

marinofaggiana (approved):

PR #1969, opened 2022-04-25
More Screens: tableviewcell mutliline text fix

bzz445:
Signed-off-by: bzz445 <bzz445@gmail.com
fix multiline
>

### DIFF
--- a/iOSClient/More/NCMore.swift
+++ b/iOSClient/More/NCMore.swift
@@ -262,7 +262,7 @@ class NCMore: UIViewController, UITableViewDelegate, UITableViewDataSource {
         if indexPath.section == 0 {
             return 100
         } else {
-            return NCGlobal.shared.heightCellSettings
+            return UITableView.automaticDimension
         }
     }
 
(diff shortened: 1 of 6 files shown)


iOSClient/More/NCMore.swift
```diff
@@ -262,7 +262,7 @@ class NCMore: UIViewController, UITableViewDelegate, UITableViewDataSource {
         if indexPath.section == 0 {
             return 100
         } else {
-            return NCGlobal.shared.heightCellSettings
+            return UITableView.automaticDimension
```
thisIsTheFoxe, line 265: This alone a) doesn't support multilines, and b) reduces the current cell height.
To make this work you would need to
1. Add: `tableView.estimatedRowHeight = NCGlobal.shared.heightCellSettings` to the `viewDidLoad`
2. (?) To keep the old height, add Top & Bottom constraints to the label, to (somewhat) keep a min. height of 50pt.

thisIsTheFoxe (commented):
Thanks also for this PR. Good idea and I can see why this is important.

However, it needs to be noted that this changes the default cell height from what it was before and I'm unsure if there is an easy solution to this for the XLForm cells (left is this PR, right is current release):

Screenshots of the settings. Now and before.
 
*Personally*, I don't think it's a big problem, especially, considering that it solves the l18n issues you mentioned. But I can also see how someone else might argue differently. @jancborchardt wdyt?

IdkSaveme (approved):

marinofaggiana:
soon a new refactoring.

Issue #1631, opened 2021-07-01
Weird text editing behavior since latest update


StCyr:
### Steps to reproduce

1. Open a text file
2. go anywhere in the text and press enter
3. the cursor jumps to the beginning of the file

### Expected behaviour

pressing enter should just go to the next line

### Actual behaviour

the cursor jumps to the beginning of the file.

Moreover, somehow, because I didn't care enough after pressing enter, the content of my file got deleted, and this empty file got automaticaly saved.

Again, for some reason, no versions of this file dating from less than 6 months are to be found in the web interface.

So, my file's content is definitively lost (well, I'll try to get it back from the backup)

### Screenshots

Here's a screenshot of my file in the web interface:

![image](https://user-images.githubusercontent.com/8179031/124186223-4b7b2300-dabc-11eb-839f-6f973fdd32d4.png)

### Environment data

**iOS version:** e.g. iOS 14.6

**Nextcloud iOS app version:** Nextcloud Liquid for iOS 4.0.1.1

**Server operating system:** Debian/stable

**Web server:** Apache

**Database:** Mysql

**PHP version:** 7.3

**Nextcloud version:**  22.0.0 RC1

StCyr:
at least I've recovered my file from the backup

StCyr:
Got it once again :-(

marinofaggiana:
Hi, text is an application, you can open here an issue :

https://github.com/nextcloud/text

StCyr:
No no, that's when I edit my file in the iOS client that the issue happens

marinofaggiana:
the editor inside ios/android use the Text app (in web interface)  as Collabora / ONLYOFFICE

timkrueger:
I reproduced the same behavior and made a short screen record.

https://user-images.githubusercontent.com/606665/128668738-c2ea981b-f6b2-4322-a274-737a43cfb224.mp4

### Environment data

iOS version: 14.6

Nextcloud iOS app version: Nextcloud Liquid for iOS 4.0.6.0

Server operating system: Synology DSM 7 (Nextcloud is running in Docker [22.0.0-apache](https://hub.docker.com/layers/nextcloud/library/nextcloud/22.0.0-apache/images/sha256-79d96e12aba703e9bed7c0deb5958db7a9682d9493efaa1cadc623213ceb62bb?context=explore))

Web server: Apache

Database: pgsql

PHP version: 7.4.21

Nextcloud version: 22.0.0

[...]

PR #1985, opened 2022-05-03
Fix user agent for text app

### LINKED ISSUE #1967: Fix the user-agent for online editing
### Steps to reproduce
1. Create blank Xcode project with a single `WKWebView`
2. Set the `WKWebView`'s`.customUserAgent` to *Mozilla/5.0 (iOS) Nextcloud-iOS/4.4.0*
3. Load a *directEditing* url with that webview
4. Try e.g. passing an image
5. Do the same but adding *AppleWebKit/605.1.15 Mobile/14E304* to the user-agent

### Actual behaviour
When loading without the additional user-agent components pasting images fails. 

### Proposed solution
Change the user-agent at least for `NCViewerNextcloudText`. 
Or if it's needed for other competent of the web app change the user-agent in general. ~~However, this would also require changes in the server.~~

Some 3rd party libraries and frameworks use the user-agent for platform specific workarounds.
This might be also be the root cause for other issues and glitches with office apps using the Nextcloud iOS app.
Would probably also be a good idea to check the Android app for similar issues @tobiasKaminsky @juliushaertl 

E.g. 3rd party libraries used by nextcloud/text:
https://github.com/ProseMirror/prosemirror-view/blob/d8c4bb1c238d814359a345e576fdfcaaa17aabd2/src/browser.js
https://github.com/ueberdosis/tiptap/blob/e07a5b625d90c77fc5820e5eefa093d3cfb96512/packages/core/src/utilities/isiOS.ts

### Logs

JS Text logs using a WKWebView, attempting to paste an image:
```
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) - http://localhost:8080/s/%7Btoken%7D/download?files=blob%3Ahttp%3A%2F%2F[...]
```

### Environment data

**iOS version:** e.g. iOS 15.4

**Nextcloud iOS app version:** `develop` branch

**Nextcloud version:** 24.0.0RC1

thisIsTheFoxe:
Fix #1967 

- allow pasting images from the pasteboard + preview (NC 24)
- fix keyboard dismisses after *Cut*
- potential fix for:
  - #1631 
  - nextcloud/text#1927

### DIFF
--- a/iOSClient/Utility/NCUtility.swift
+++ b/iOSClient/Utility/NCUtility.swift
@@ -28,6 +28,7 @@ import NCCommunication
 import PDFKit
 import Accelerate
 import CoreMedia
+import WebKit
 
 // MARK: - NCUtility
 
@@ -303,6 +304,19 @@ class NCUtility: NSObject {
             }
         }
         return true
+    }
+
+    @objc func getCustomUserAgentNCText() -> String {
+        let userAgent: String = CCUtility.getUserAgent()
+        if UIDevice.current.userInterfaceIdiom == .phone {
+            // NOTE: Hardcoded (May 2022)
+            // Tested for iPhone SE (1st), iOS 12; iPhone Pro Max, iOS 15.4; iPhone X,
+            // 605.1.15 = WebKit build version
+            // 15E148 = frozen iOS build number according to: https://chromestatus.com/feature/4558585463832576
+            return userAgent + " " + "AppleWebKit/605.1.15 Mobile/15E148"
+        } else {
+            return userAgent
+        }
     }
 
     @objc func getCustomUserAgentOnlyOffice() -> String {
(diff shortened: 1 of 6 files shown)


iOSClient/Utility/NCUtility.swift
```diff
@@ -28,6 +28,7 @@ import NCCommunication
 import PDFKit
 import Accelerate
 import CoreMedia
+import WebKit
```
marinofaggiana, line 31: Need WebKit ?

marinofaggiana (changes requested):
Read comments.

iOSClient/Utility/NCUtility.swift
thisIsTheFoxe: Used for testing. Removed it.

marinofaggiana (approved):

PR #1987, opened 2022-05-04
Pdfview Improvements

marinofaggiana:
- Open / Close thumbnails with the pdf zoomed
- Remove image
- Added EasyTipView in Acknowledgements


thisIsTheFoxe (approved):
lgtm, works.. 👍 

I also noticed that on iOS 12 the PDFView has a transparent background which then is also used for the thumbnails.. idk, maybe you could only change the bg for iOS 13+, else use the default?

Screenshot 2022-05-05 at 09 11 40

PR #1994, opened 2022-05-09
Fix passcode is not presented

### LINKED ISSUE #1992: Touch ID doesn't work and NC crashes
### Steps to reproduce

Hello Marino, 

Since version 4.4.0, Touch ID doesn't work neither on my iPhone 8 Plus nor on my iPad Air 3. 

The screen shows a numeric keypad and "Touch ID", but tapping on it is ineffective. 
I checked the settings, the lock code and Touch ID are enabled. And despite restarting the iPhone, Touch ID does not work. 
Touch ID works with version 4.3.0 (4). 

Also, Nextcloud crashes in the background. After a crash, NC does not ask for the lock password again.

Lise

220508_iPhone8505

### Environment data

**iOS version:** e.g. iOS 15.4.1

**Nextcloud iOS app version:** 4.4.0.8

**Nextcloud server version:** 23.0.2

thisIsTheFoxe:
Fix #1992 
It's a bug introduced in #1964 since the passcode is not presented on the main window.

### DIFF
--- a/iOSClient/AppDelegate.swift
+++ b/iOSClient/AppDelegate.swift
@@ -727,30 +727,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             completion()
         })
     }
-    
-    func isPasscodePresented() -> Bool {
-        return window?.rootViewController?.presentedViewController is TOPasscodeViewController
-    }
-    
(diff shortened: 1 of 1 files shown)


marinofaggiana:
works.

marinofaggiana (approved):

PR #1996, opened 2022-05-09
TestFlight

marinofaggiana:
On settings added the TestFlight option


thisIsTheFoxe (changes requested):
Hmm.. Nice idea, but:
1. _View in App Store_ doesn't work well (should open iOS App Store):
IMG_FBAB5D66B328-1
2. _Start Testing_ doesn't work at all (nothing happens)
3. It would be cool to directly open TestFlight deep linked, if it's installed

PR #2148, opened 2022-09-10
l10n: Improved grammar

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>

### DIFF
--- a/iOSClient/Supporting Files/en.lproj/Localizable.strings
+++ b/iOSClient/Supporting Files/en.lproj/Localizable.strings
@@ -588,7 +588,7 @@
 "_password_pdf_error_"          = "Wrong password";
 "_error_download_photobrowser_" = "Error: Unable to download photo";
 "_good_morning_"                = "Good morning";
-"_good_noon_"                   = "Good noon";
+"_good_noon_"                   = "Good day";
 "_good_afternoon_"              = "Good afternoon";
 "_good_evening_"                = "Good evening";
 "_good_night_"                  = "Good night";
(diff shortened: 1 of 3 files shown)


iOSClient/Supporting Files/en.lproj/Localizable.strings
```diff
@@ -588,7 +588,7 @@
 "_password_pdf_error_"          = "Wrong password";
 "_error_download_photobrowser_" = "Error: Unable to download photo";
 "_good_morning_"                = "Good morning";
-"_good_noon_"                   = "Good noon";
+"_good_noon_"                   = "Good day";
```
Valdnet, line 591: Why "Good day" if the text string appears exactly at 12 o'clock?
https://github.com/nextcloud/ios/blob/17443651c8b978eef74b046241a72ebd936d5476/Widget/Nextcloud/NextcloudData.swift#L62

Valdnet:
I would not use this phrase at present. Immediately after 12 o'clock I would give "Good afternoon".

But if we want to highlight a greeting between 12:00 to 12:59, "Good day" is actually the more popular 👍.

To keep the code tidy, I would change this value: "_good_noon_" to "_good_day_".

Valdnet (approved):

marinofaggiana:
Changed in hardcode. thanks

marinofaggiana:
@rakekniven 👍🏻 

- "_good_noon_"                   = "Good noon";
+ "_good_noon_"                   = "Good day";

PR #2228, opened 2022-10-30
Ensure that maximum expire dates are enforced when the server enforces them

claucambra:
This makes the iOS client follow the server settings on enforced expire days while preventing errors when sending expire dates that are not allowed.

Apologies if the code is not very Swift-y, I'm a lot more accustomed to Obj-C 🙂

Signed-off-by: Claudio Cambra <developer@claudiocambra.com>


marinofaggiana:
Thanks @claucambra !

PR #2354, opened 2023-02-13
Adapt to rework of "shared" design of NKCommon

claucambra:
Counterpart to https://github.com/nextcloud/NextcloudKit/pull/19


marinofaggiana:
@claucambra no problem I create the "Adapt to rework of "shared" design of NKCommon" soon so I change other things.

PR #2360, opened 2023-02-22
fix: Open links from text in Safari

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>


marinofaggiana:
I don't have tested but the code it's ok.

marinofaggiana:
Tested: works.

PR #2420, opened 2023-05-16
CI test integration

mpivchev:
This adds integration for Unit, Integration, UI, and End to End tests


.github/workflows/xcode.yml
```diff
@@ -37,7 +38,7 @@ jobs:
       run: wget "https://raw.githubusercontent.com/firebase/quickstart-ios/master/mock-GoogleService-Info.plist" -O GoogleService-Info.plist
     - name: Build & Test Nextcloud iOS
       run: |
-        xcodebuild clean build test -project $PROJECT -scheme "$SCHEME" -destination "$DESTINATION"
+        xcodebuild clean build test -quiet -project $PROJECT -scheme "$SCHEME" -destination "$DESTINATION"
```
mpivchev, line 41: @marinofaggiana do we need to `clean` every time? It only adds to the build time

PR #2430, opened 2023-05-20
Capabilitiesview

marinofaggiana:
New Capabilities view

### DIFF
--- a/iOSClient/Diagnostics/NCCapabilitiesView.swift
+++ b/iOSClient/Diagnostics/NCCapabilitiesView.swift
@@ -31,7 +31,7 @@ class NCCapabilitiesViewOO: ObservableObject {
     }
 
     @Published var capabililies: [Capability] = []
-    @Published var json = "Lorem ipsum dolor sit amet. Ea voluptas aperiam aut inventore saepe in tenetur modi\n.Cum sint tempore sed maiores quos aut quaerat deleniti.\nQui beatae quia qui repellat sunt in Quis libero aut quidem porro non explicabo tenetur et natus doloribus non voluptatum consequatur.\n"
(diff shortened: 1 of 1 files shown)


iOSClient/Diagnostics/NCCapabilitiesView.swift
```diff
@@ -0,0 +1,236 @@
...
+
+        // GROUP FOLDERS
+        available = NCManageDatabase.shared.getCapabilitiesServerBool(account: account, elements: NCElementsJSON.shared.capabilitiesGroupfoldersEnabled, exists: false)
+        capabililies.append(Capability(text: "Group folders", image: UIImage(systemName: "person.2")!, available: available))
+
+        if let json = NCManageDatabase.shared.getCapabilities(account: account) {
+            self.json = json
+        }
+    }
+}
```
mpivchev, line 162: Is there a reason we don't put all this business logic into view models? The view controller becomes really big.

iOSClient/Diagnostics/NCCapabilitiesView.swift
```diff
@@ -0,0 +1,236 @@
...
+        if let editors = NCManageDatabase.shared.getDirectEditingEditors(account: account) {
+            for editor in editors {
+                if editor.editor == NCGlobal.shared.editorText {
+                    textEditor = true
+                } else if editor.editor == NCGlobal.shared.editorOnlyoffice {
+                    onlyofficeEditors = true
+                }
+            }
+        }
+        capabililies.append(Capability(text: "Text", image: UIImage(named: "text")!.resizeImage(size: CGSize(width: 25, height: 25))!, available: textEditor))
```
mpivchev, line 126: Lots of force unwrapping here. Can we use `guard`?

mpivchev: You are mostly right, but there are edge cases where you might rename an asset or remove it and the whole app crashes because of a random image. I still think guarding is a good idea here but up to you.

iOSClient/Diagnostics/NCCapabilitiesView.swift
mpivchev: In any case it's probably too late to make any kind of a big structural change in the project

mpivchev (approved):

PR #2439, opened 2023-05-25
account checked on delete

marinofaggiana:
- a little forgetfulness can prove fatal error

### DIFF
--- a/iOSClient/Menu/NCMenuAction.swift
+++ b/iOSClient/Menu/NCMenuAction.swift
@@ -139,6 +139,7 @@ extension NCMenuAction {
                         Task {
                             var error = NKError()
                             var ocId: [String] = []
+                            let account = selectedMetadatas.first?.account ?? ""
                             for metadata in selectedMetadatas where error == .success {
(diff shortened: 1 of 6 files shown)


iOSClient/Menu/NCMenuAction.swift
```diff
@@ -169,7 +171,7 @@ extension NCMenuAction {
                             if error != .success {
                                 NCContentPresenter.shared.showError(error: error)
                             }
-                            NotificationCenter.default.postOnMainThread(name: NCGlobal.shared.notificationCenterDeleteFile, userInfo: ["ocId": ocId, "error": error])
+                            NotificationCenter.default.postOnMainThread(name: NCGlobal.shared.notificationCenterDeleteFile, userInfo: ["account": account, "ocId": ocId, "error": error])
```
mpivchev, line 174: We are repeating this code twice. Can we shorten it?

mpivchev (changes requested):

mpivchev (approved):
Approved, but we should strive for improving our code where we can, otherwise code debt will increase

PR #2440, opened 2023-05-26
Fix - Improvements

marinofaggiana:
- Removed TagListView from cell (unnecessary)
- Fix scroll in collection common
- Set version 4.8.2
- Set min available version iOS to 15
- Capabilities refactoring


mpivchev:
Build fails

mpivchev:
These are the old tests we had for this app. You can delete them.

iOSClient/Data/NCManageDatabase+Capabilities.swift
```diff
@@ -0,0 +1,136 @@
+//
```
mpivchev, line 1: This should be in `iOSClient/Extensions`

PR #2458, opened 2023-06-09
Add unique string to shared file names

mpivchev:
Some files that are shared from other apps, like WhatsApp, might have the same names. For example photos that were captured at the same minute. To fix this, I added a unique string for those file names so they are always different.

Before the fix some files that are different photos would sometimes have the same name, like the 3 pictures captured at the same minute:
```
Name Optional("PHOTO-2023-03-31-20-46-01.jpg")
Name Optional("PHOTO-2023-03-31-20-46-01.jpg")
Name Optional("PHOTO-2023-03-31-20-46-01.jpg")
Name Optional("PHOTO-2023-04-04-19-11-44.jpg")
Name Optional("PHOTO-2023-04-04-19-12-04.jpg")
Name Optional("PHOTO-2023-04-07-11-41-11.jpg")
```

This would cause the app to ignore the files with the same name and the loaded files to be less than the selected ones (6 selected but 4 loaded):

image

The fix adds a unique string:
```
Name Optional("PHOTO-2023-03-31-20-46-010054.jpg")
Name Optional("PHOTO-2023-03-31-20-46-010055.jpg")
Name Optional("PHOTO-2023-03-31-20-46-010056.jpg")
Name Optional("PHOTO-2023-04-04-19-11-440057.jpg")
Name Optional("PHOTO-2023-04-04-19-12-040058.jpg")
Name Optional("PHOTO-2023-04-07-11-41-110059.jpg")
```

Now loads all files, even the ones that are different photos but have the same name:

image

### DIFF
--- a/Share/NCShareExtension+Files.swift
+++ b/Share/NCShareExtension+Files.swift
@@ -115,7 +115,7 @@ class NCFilesExtensionHandler {
                 var originalName = (dateFormatter.string(from: Date())) + String(ix)
 
                 if let url = item as? URL, url.isFileURL, !url.lastPathComponent.isEmpty {
-                    originalName = url.lastPathComponent
+                    originalName = url.deletingPathExtension().lastPathComponent + CCUtility.getIncrementalNumber() + "." + url.pathExtension
                 }
 
                 var fileName: String?
@@ -131,7 +131,9 @@ class NCFilesExtensionHandler {
                 default: return
                 }
 
-                if let fileName = fileName, !filesName.contains(fileName) { filesName.append(fileName) }
+                if let fileName, !filesName.contains(fileName) {
+                    filesName.append(fileName)
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


mpivchev:
Yes, this can be a problem.
 I can try to get the seconds, though I'm not sure it's possible since it's not a `PHAsset`. 
Or I can only do this if I find out there are multiple files that are selected with the same name?

mpivchev:
I tried iCloud and there they add a number to the files with the same name, and obviously they don't check if the file already exists. We can leave that to the user.

mpivchev:
Now it only adds a unique string to files with the same name:

image

The rest of the files are not touched:

image

This is similar to how iCloud does it.

Share/NCShareExtension+Files.swift
```diff
@@ -102,20 +103,24 @@ class NCFilesExtensionHandler {
         var counter = 0
 
         self.itemsProvider = items.compactMap({ $0.attachments }).flatMap { $0.filter({
-            $0.hasItemConformingToTypeIdentifier(kUTTypeItem as String) || $0.hasItemConformingToTypeIdentifier("public.url")
+            $0.hasItemConformingToTypeIdentifier(UTType.item.identifier as String) || $0.hasItemConformingToTypeIdentifier("public.url")
```
mpivchev, line 106: Above was deprecated on iOS 15+ so I changed it to new implementation

Issue #2460, opened 2023-06-12
Add other Nextcloud apps to settings as suggestions


mpivchev:
iOS Files:
- in settings: above "Transfers" with gap: Talk, Notes, "more apps"
- use same "more" icon as Desktop
- opens app if installed, if not open link to AppStore
- "more" -> open Appstore with filter for "Nextcloud"
- with multi account: open another app with correct account

_Originally posted by @tobiasKaminsky in https://github.com/nextcloud/files-clients/issues/10#issuecomment-1560748782_

Update:

- [ ] When opening another app from the Files app we should switch to the same account in the other app. This should only happen if that account is logged in the other app, otherwise we just open the app as if opening it from the app drawer.
- [ ] Notes app does not have multi-account support, or a deeplink URL. https://github.com/nextcloud/notes-ios/issues/40
- [ ] Talk app has no deeplink URL for switching accounts when opening app. https://github.com/nextcloud/talk-ios/issues/1286
- [x] Files app for now has implemented switching apps, but without account switching: https://github.com/nextcloud/ios/pull/2461

Until the above are finished, we will just open the apps as if opening them from the file drawer. No account switching or other functionality.

mpivchev:
@tobiasKaminsky Not sure where to find this "More" icon. I can't find it on the site

mpivchev:
This is rolled back for now since a bug was found where "Settings" would not show among other things (oops)

mpivchev:
@marinofaggiana you mentioned you are not a fan of the design. Maybe we can discuss it here? @jancborchardt 

Current design: 

<img src="https://github.com/nextcloud/ios/assets/6960329/83b85ed5-3884-4aba-8c7f-8668dd724733" width="300" height="650">

mpivchev:
Hmm, personally I'm not a fan of hamburger menus and they are not a native iOS component. No official iOS app uses them, and I've noticed apps that did use them slowly transitioned to removing them and using a tab bar instead, just as how we are now.

mpivchev:
We can instead have an account tab with all the most useful and needed settings, and have a separate settings icon on the top right for more advanced settings that people don't use often. This is how Dropbox does it:

![image](https://github.com/nextcloud/ios/assets/6960329/6db60172-110f-42d4-a910-86fe2d74d392)

mpivchev:
If we do decide on using a hamburger icon, then we can replace the More tab with a Shared tab instead, so users can quickly go to that instead of through settings as they do now

[...]

PR #2461, opened 2023-06-12
Add new section in settings for Nextcloud apps

mpivchev:
This adds a new section in settings to promote our other apps:

- If the app is installed, it opens it without doing anything else. 

    - Note: Originally it was supposed to open the app and switch accounts (if account is logged in), but that is not supported by Talk or Notes yet. Separate issues have been made. See: https://github.com/nextcloud/ios/issues/2460. For now I suggest we do it this way and move this PR forward until we have the other issues done @tobiasKaminsky @jancborchardt 

- If the app is not installed, it leads to the app store page.
- If the "more apps" icon is clicked, it opens a web page with other Nextcloud apps.

Additionally, a new URL scheme was made to add support for other apps to open this app and switch accounts (if account is logged in):

```
nextcloud://open-and-switch-account?user=someUser&url=https://cloud.nextcloud.com
```

Finally, refactored some of the code and removed duplicate code.


iOSClient/More/NCMore.swift
```diff
@@ -483,53 +498,8 @@ class NCMore: UIViewController, UITableViewDelegate, UITableViewDataSource {
...
-            frame.size.width = newWidth
-            frame.origin.x += space
-            super.frame = frame
-        }
-    }
-}
-
-class NCMoreUserCell: UITableViewCell {
-
-    @IBOutlet weak var displayName: UILabel!
```
mpivchev, line 517: These were separated in their own files

iOSClient/More/Cells/BaseNCMoreCell.swift
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+
+        selectedBackgroundView = selectionColor
+        backgroundColor = .secondarySystemGroupedBackground
+        layer.cornerRadius = defaultCornerRadius
+    }
```
mpivchev, line 35: This code was used for every cell, so I made a base class instead.

tobiasKaminsky:
Can you use 
[image](https://github.com/nextcloud/desktop/blob/d8479fb1bafacbef2ecdb26516b9c7f513204f01/theme/white/more-apps.svg?plain=1)

for more icon?

mpivchev:
> we could take the opportunity to migrate this simple view into swiftUI and the GUI does not convince me

Hmm, we could've decided to do this earlier before these changes were made. But I can try nonetheless.

> the GUI does not convince me

Any suggestions are welcome. We can maybe ask some of our UI designers to design a view.

jancborchardt (changes requested):
Nice! Design-wise, how about something standard like Apple also uses on the top of Contacts? That has proper spacing and all.

mpivchev:
@jancborchardt you mean this part?

image

Since we only have 3 of these, should they be stacked left or be in the middle?

marcoambrosini (commented):
I don't think that these shortcuts should be the first entry in the settings because they can be detrimental to usability and create friction. I would say that settings is probably the wrong place to advertise the presence of other apps, and I suggest something subtler and "dismissable". if we absolutely have to, what about moving them to the bottom of the list?
Also, would these be visible even for users that have these apps installed?

mpivchev:
> I don't think that these shortcuts should be the first entry in the settings because they can be detrimental to usability and create friction. I would say that settings is probably the wrong place to advertise the presence of other apps, and I suggest something subtler and "dismissable". if we absolutely have to, what about moving them to the bottom of the list?

I see your point. I still need to move this forward however, so if @jancborchardt is ok with this I can move this section to the bottom.

jancborchardt:
> Since we only have 3 of these, should they be stacked left or be centered?

@mpivchev is there any recommended default or standard? Centered seems good.

---

@marcoambrosini the issue if they are on the bottom is that they will be difficult to discover, defeating the purpose. It is obvious that the vertical list of entries goes on, but it would not be obvious that there is an entirely new element below.

mpivchev:
@jancborchardt Here is the updated look:
image

mpivchev:
@jancborchardt Here is the new look:

jancborchardt:
In iOS Contacts, there is a bit more top and bottom padding inside the entries. Can you adjust that @mpivchev? :)

mpivchev:
@jancborchardt Is this better?

mpivchev:
@jancborchardt pinging again in case you missed it

jancborchardt (approved):
Latest screenshot looks good! :)

hannesfritz:
iOS seems to stretch those buttons if less are available. I just found an example with 4 instead of 5 like in the screenshot above. But I did an additional mockup to show what it might look with only the three options. 
In my opinion it is calmer on the eye, better aligned and additionally larger tap targets.

So just food for thought :) 
Contacts 4 jumps
Nextcloud more

jancborchardt:
Yeah, that seems very nice @hannesfritz!

@mpivchev could you adjust them like that? :)

mpivchev:
Branch restored so I can reimplement this soon

Issue #2490, opened 2023-06-26
Set default chunk_size != 0


tanganellilore:
Hi team,

I notice that in Iphone app, the default chunk_size is set to 0, and max value is equal to 100 (in MB).
To align this to desktop client and andorid up, I think it's bettere to set it at 100MB as per default and then allow all type of value (>100MB).

I will try to send you a PR to fix it.

PR #2491, opened 2023-06-26
Set default and max value to chunkSize

tanganellilore:
* Resolve: #2490 

Hi team,
as described on issue above, I set default value for chunks to 100MB and increase the max limit to 1000MB.

Actually is not see any "flow" to check the best chunksize with nextcloud server, but this minor fix, for sure help a lot all people that have timeout issue or limit on nextcloud server.

I can increate max value for chunks, but to me is not much sense (is a mobile, so not much usefull).

Let me know if it's ok for you.

### DIFF
--- a/iOSClient/Settings/CCAdvanced.m
+++ b/iOSClient/Settings/CCAdvanced.m
@@ -116,7 +116,7 @@
     [row.cellConfigAtConfigure setObject:@YES forKey:@"stepControl.wraps"];
     [row.cellConfigAtConfigure setObject:@1 forKey:@"stepControl.stepValue"];
     [row.cellConfigAtConfigure setObject:@0 forKey:@"stepControl.minimumValue"];
-    [row.cellConfigAtConfigure setObject:@100 forKey:@"stepControl.maximumValue"];
+    [row.cellConfigAtConfigure setObject:@1000 forKey:@"stepControl.maximumValue"];
     [section addFormRow:row];
 
     // Section : Privacy --------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mpivchev:
We are currently working on optimizations regarding our chunked upload, see: https://github.com/nextcloud/files-clients/issues/28

Once this is tested and done we will review this PR.

tanganellilore:
One point, setting `chunkSize `to 0 is equal to NOT chunk.... so in case of i tryt o upload a 1-2GB video file to Nextcloud, and my limit is set to 512MB, I will receive errors.

tanganellilore:
So this means that if you upload a 1gb of 4k video, upload file  will fail, due to not chunks as per default.

To me is not the correct approach

tanganellilore:
Why this was closed?

Gymnae:
+1 for reopen and introducing a proper method. 
Why is chunked uploading only possible if the app is in the foreground?

juliushaertl:
From the server side using chunked uploading is much more reliable for larger files, so I'd agree that this should be re-evaluated on the client side. @marinofaggiana Can you elaborate a few more technical insights on why this doesn't work? Is there a way this could be implemented in a more reliable way?

This would also be relevant for https://github.com/nextcloud/files-clients/issues/28 as otherwise this optimization doesn't add much benefit (cc @tobiasKaminsky)

tanganellilore:
Hi @marinofaggiana , thanks for explanation but i have some concern on it.

What is the meaning of "some large file, for small and 'normal' files only ever use foreground mode":
"large file" it's so generic, for my understand this app will not differentiate small file to big file.

Reapeat 0 = unlimited, so it's better start upload 4GB of file and receive an error when we reach the nextcloud server upload limit (like 1GB) and restart the upload again?

It's possible think/elaborate a strategy to send all "small" file (user can be decide what is small for him) on foreground and send large file with chunked in the background 8as per best practive suggested by apple)?


Thanks

PR #2493, opened 2023-06-27
Attempt to fix selecting folders in iOS (#943)

### LINKED ISSUE #943: Can't add Nextcloud folders in other apps
### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             

Add Nextcloud folders in other apps, eg. in Textastic. There are only a few complete first class file storage citizen on iOS who supports this feature, eg. Working Copy or Secure Shell Fish

This is what I mean:
https://www.textasticapp.com/v8/manual/integration_other_apps/external_files_folders.html

### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               

Can't add Nextcloud folders in Textastic or 1Writer - Nextcloud storage is just greyed out.

### Steps to reproduce                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             

Open Textastic or 1Writer (or any other app which supports adding folders) and try it with Nextcloud storage provider.

### iOS version

13.1 (but it never worked, even with iOS 12)

### App version

2.24.1 (also never worked before)

### Server configuration

Doesn't matter, it's an iOS topic

thomwiggers:
This is an attempt to fix #943. However, _I cannot test this_, so I would appreciate any help.

This is based on https://twitter.com/benoitsan/status/1380507723020783619


mpivchev:
Hi, thanks for your PR. I will check it out soon. I see that the commits are not signed so the DCO check fails. Please amend your commit so it's signed. More info here: https://github.com/nextcloud/ios#how-to-sign-off

thomwiggers:
I've amended the commit

thomwiggers:
The intended aim of this PR is to add support for tools like Logseq. So hopefully, after merging, the following should work:

1. Download the Logseq app
2. Try to open a Nextcloud folder as the source of the notes.

Unfortunately, I do not have a Mac so I can not build the app or install development versions of apps.

thomwiggers:
I also do not know if more is needed than just specifying this key in the app's capabilities, so this should definitely be tested.

mpivchev:
I'll try testing this

mpivchev:
Upon trying to access and open a folder in the Nextcloud directory, I get this:

IMG_2729

IMG_2730

So, either our app folder is not ready to be accessed by other apps yet, or this Logseq app does not support opening other app folders, or both.

thomwiggers:
Shame, seems it is a little bit harder than just adding this entry to the manifest. I don't immediately know of another app that could be used to test this; but I would not be surprised if indeed a bit more needs to be done on the nextcloud side. Thanks for trying it out.

I'm closing this PR, as I don't know how to develop this further, nor do I have the resources (a mac) to attempt anything.

miguelarios:
@thomwiggers the only other app that I know if Cryptomator

Nebucatnetzer:
There are plenty of app to test it.
E.g. with Shortcuts you have a builtin app that can’t access the Nextcloud folder.
Textastic is another one.

PR #2494, opened 2023-06-27
Remove command buttons

mpivchev:
This removes the top command buttons as they are redundant.

Before and after:

-------
Also changed sorting to use iOS native icons.

Before:

image
image

After:

image
image


jancborchardt (approved):
Looks good! :)

PR #2496, opened 2023-06-28
Hide "show more results" text when there are no more files to display in search

### LINKED ISSUE #2495: "Show more results" text should not display when there are no more files to display in Search
Support ticket: https://support.nextcloud.com/#ticket/zoom/47469

The button should stop showing when there are no more files, worst case scenario at least after one more press and a second request.

https://github.com/nextcloud/ios/assets/6960329/78572ab4-4cba-4222-8fab-12df59b1f996

mpivchev:
This will close #2495 

I added this based on what the API currently provides.

Since there is no way to check the total entries or if it's the last page through the API, I check it locally when searching for more entries.

Suggestions are welcome if there is a better way.

Here is how it looks like:

https://github.com/nextcloud/ios/assets/6960329/d314a204-cd65-4ed6-8350-cf29754aa30b


mpivchev:
> Works but not good ... you can added a new check if results < limit: (5) then isLastPage = true
> 
> 
> 
> look the video
> 
> 
> 
> https://github.com/nextcloud/ios/assets/8616947/e984e1fc-93ab-4f72-8d31-f30fc04c8507
> 
> 
> 
> 
> 
> 

Sure, but when we call the API for the first time we don't set a limit, so the limit is decided by the server. If the server someday decides to change the limit then the app might skip some search results.

iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
```diff
@@ -1736,12 +1732,19 @@ extension NCCollectionViewCommon: UICollectionViewDataSource {
...
                 if sections > 1 && section != sections - 1 {
                     footer.separatorIsHidden(false)
                 }
 
-                if isSearchingMode && isPaginated && metadatasCount > 0 && !isLastPage {
+                // If the number of entries(metadatas) is lower than the cursor, then there are no more entries.
+                // The blind spot in this is when the number of entries is the same as the cursor. If so, we don't have a way of knowing if there are no more entries.
+                // This is as good as it gets for determining last page without server-side flag.
+                let isLastPageTest = (metadatasCount < metadataForSection?.lastSearchResult?.cursor ?? 0) || metadataForSection?.lastSearchResult?.entries.isEmpty == true
+
```
mpivchev, line 1745: Important to note

mpivchev:
@marinofaggiana I will ask server-side if they can add some kind of flag for this

iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
mpivchev: `isLastPageTest` was renamed to `isLastPage`, besides that ignore the Outdated tag

mpivchev:
Also added integration test: https://github.com/nextcloud/NextcloudKit/pull/35

mpivchev:
@marinofaggiana Let's merge this fix for now until hopefully there is a server fix for this at some point?

PR #2512, opened 2023-07-05
Collection view common fix + lint Data

marinofaggiana:
+ lint
+ fix on NCCollectionViewCommon

### DIFF
--- a/iOSClient/Data/NCManageDatabase+Metadata.swift
+++ b/iOSClient/Data/NCManageDatabase+Metadata.swift
@@ -672,13 +672,12 @@ extension NCManageDatabase {
 
     func setMetadataEtagResource(ocId: String, etagResource: String?) {
 
-        var result: tableMetadata?
         guard let etagResource = etagResource else { return }
 
         do {
             let realm = try Realm()
             try realm.write {
-                result = realm.objects(tableMetadata.self).filter("ocId == %@", ocId).first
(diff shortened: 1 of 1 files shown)


iOSClient/Data/NCManageDatabase+Account.swift
```diff
@@ -262,22 +264,24 @@ extension NCManageDatabase {
...
+        do {
+            let realm = try Realm()
+            guard let result = realm.objects(tableAccount.self).filter("active == true").first else { return 1 }
+            return result.autoUploadSubfolderGranularity
+        } catch let error as NSError {
+            NextcloudKit.shared.nkCommonInstance.writeLog("Could not write to database: \(error)")
         }
 
-        return result.autoUploadSubfolderGranularity
+        return 1
```
mpivchev, line 275: Can we extract these magic 1s into a variable or a constant?

mpivchev (changes requested):

mpivchev:
Can you specify in more detail what exactly this is fixing? "fix on NCCollectionViewCommon" doesn't really say much, thanks.

iOSClient/Data/NCManageDatabase+Account.swift
mpivchev: This was not resolved

Issue #2425, opened 2023-05-17
Missing sharing options


freumichCH:
### Steps to reproduce

1. Create internal share for user or group over webGUI
2. Open sharing options for previously created share in iOS App

### Expected behaviour

Sharing options on Nextcloud iOS App should show the same option set like WebGUI. 

### Actual behaviour

Nextcloud iOS shows only... 
- Allow resharing 
- Allow editing
- Allow creating
- Allow deleting

but not....
- Allow download

### Screenshots
<img width="572" alt="signal-2023-05-17-140110" src="https://github.com/nextcloud/ios/assets/22443781/939de7d8-f0c8-4a3c-b887-1ff770ff05a2">

![Untitled(1)](https://github.com/nextcloud/ios/assets/22443781/000da2c7-9134-43c2-b52a-11f8a4c7b714)

**iOS version:** 
iOS 16.4.1

**Nextcloud iOS app version:** see More > Settings
4.8.0.5

**Server operating system:**
Ubuntu

**Web server:** Apache, nginx
Apache

**Database:**
mySQL

**PHP version:**
7.4

**Nextcloud version:** 
24.0.12

PR #2535, opened 2023-07-16
Sharing attributes - Allow download folder

marinofaggiana:
https://github.com/nextcloud/ios/issues/2425


iOSClient/Data/NCManageDatabase+Share.swift
```diff
@@ -236,4 +238,32 @@ extension NCManageDatabase {
...
+                        let scope = sub["scope"] as? String
+                        if key == "download", scope == "permissions", let enabled = enabled {
+                            return enabled
+                        }
+                    }
+                }
+            } catch let error as NSError { print(error) }
+        }
+        return true
+    }
```
mpivchev, line 267: Can't we use SwiftyJson for this?

mpivchev (changes requested):
Looks good! Just a few things

iOSClient/Data/NCManageDatabase+Share.swift
mpivchev: Cool, I guess Swift has built in functions for all it does now.

Issue #2532, opened 2023-07-14
Upload file -> selecting multiple files -> only 1 file uploads


mpivchev:
https://github.com/nextcloud/ios/assets/6960329/233acd2c-747b-40e0-9f35-1fd430a32371

PR #2538, opened 2023-07-17
Add unit tests for some realm functions

mpivchev:
Tried to add some tests for #2532. Couldn't test as much as I thought (we use logic from iOS files picker which is UI), but i was still able to test if realm saves the upload process data properly.


Tests/NextcloudUnitTests/DatabaseUnitTests.swift
```diff
@@ -0,0 +1,60 @@
...
+ */
+class NextcloudUnitTests: XCTestCase {
+    var realm: Realm!
+
+    private let file1Name = "file-sample1.pdf"
+    private let file2Name = "file-sample2.pdf"
+    private let file3Name = "file-sample3.pdf"
+
+    override func setUp() {
+        Realm.Configuration.defaultConfiguration.inMemoryIdentifier = self.name
```
mpivchev, line 25: Creating an in-memory Realm instance, used in the whole app during testing.

mpivchev:
The tests fail because of some architecture issue -_- Might open an issue with Realm for this. Unless you know why it may be happening? @marinofaggiana

PR #2539, opened 2023-07-18
improvements on Capabilities data

marinofaggiana:
- Capabilities have grown enormously and the SwiftyJSON proves terribly slow.
- remove the SwiftyJSON and use the swift Codable JSONDecoder


mpivchev (approved):
Nice! I guess we still need SwiftyJSON for other parts?

PR #2541, opened 2023-07-19
Use multipartupload for uploading chunks to s3

mpivchev:
Docs: https://github.com/nextcloud/documentation/pull/10740
Dependency PR: https://github.com/nextcloud/NextcloudKit/pull/38


mpivchev:
@marinofaggiana @tobiasKaminsky  Why do we need to specify the chunk size manually inside the apps? On web chunking happens automatically.

Also, you have to adjust it constantly if you upload different files. For example I wanted to upload a 20 MB file so I set the chunks to 5MBs, but then I wanted to upload a 20 GB file and I wanted to change them to 1GB. Having this done automatically would be a much better experience. We can still keep the option for manual chunk size as well. 

@marinofaggiana Second thing, the docs specify that the min chunk is 5MB and the max can be 5GB. Currently, the max in the app is 1GB. Can we increase it?

mpivchev:
What else to add:

- [ ] Rounding 5MB is not the same as what the server wants, so we get an error. Need to round it properly.
- [ ] Change the chunk size UI element to have 5MB as min size.

iOSClient/Utility/CCUtility.m
```diff
@@ -678,8 +679,11 @@ + (NSInteger)getChunkSize
     }
 }
 
+/// In megabytes (MB)
 + (void)setChunkSize:(NSInteger)size
 {
+    if (size < 10) size = 0;
```
juliushaertl, line 685: I think the size should still be default to the minimum supported value of 5MB then if configured to be > 0.

Maybe there is some byte calculation in another place where the 5 MB -> bytes generates the wrong value?

juliushaertl (approved):
Small comment left regarding the chunk size, but tested and works and also verified that the S3 upload mechanism is properly used.

iOSClient/Utility/CCUtility.m
mpivchev: You mentioned the min server size is 5 MiB, not MB. So maybe we will set it to this, or keep it to 10MB as minimum. We are also adding auto chunking, so we'll see how we'll implement this

iOSClient/Utility/CCUtility.m
```diff
@@ -667,19 +667,24 @@ + (void)setAccountRequest:(BOOL)set
...
         return 0;
     } else {
         return [size integerValue];
     }
 }
 
+/// In megabytes (MB)
 + (void)setChunkSize:(NSInteger)size
 {
+    if (size < 10) size = 0;
```
mpivchev, line 686: Int checks are temporary until we add auto chunking

mpivchev:
This will close https://github.com/nextcloud/ios/issues/2471

PR #2546, opened 2023-07-24
GUI upload (foreground)

marinofaggiana:
- Totally changed the GUI for uploading files E2EE and CHUNK which they require the foreground mode
- Some fix detect in the chunk upload

### DIFF
--- a/iOSClient/Data/NCDataSource.swift
+++ b/iOSClient/Data/NCDataSource.swift
@@ -219,79 +219,6 @@ class NCDataSource: NSObject {
     }
 
     @discardableResult
-    func addMetadata(_ metadata: tableMetadata) -> (indexPath: IndexPath?, sameSections: Bool) {
-
-        let numberOfSections = self.numberOfSections()
-        let sectionValue = getSectionValue(metadata: metadata)
-
-        // ADD metadatasSource
-        if let rowIndex = self.metadatas.firstIndex(where: {$0.fileNameView == metadata.fileNameView || $0.ocId == metadata.ocId}) {
-            self.metadatas[rowIndex] = metadata
-        } else {
-            self.metadatas.append(metadata)
-        }
-
-        // ADD metadataForSection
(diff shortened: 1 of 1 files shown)


iOSClient/Data/NCManageDatabase+Metadata.swift
```diff
@@ -867,9 +874,9 @@ extension NCManageDatabase {
 
     func getMetadataFromOcId(_ ocId: String?) -> tableMetadata? {
 
+        guard let ocId = ocId else { return nil }
```
mpivchev, line 877: you can just do `guard let ocId`

mpivchev (changes requested):
image

- String seems broken
- Top margin is bigger than bottom
- Pull to refresh causes progress bar to flicker
- Switching tabs causes progress bar to flicker

mpivchev:
Tests also failing, use develop for NextcloudKit

mpivchev:
DCO also fails

mpivchev:
@marinofaggiana rebase develop to resume tests

iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
```diff
@@ -544,11 +544,14 @@ class NCCollectionViewCommon: UIViewController, UIGestureRecognizerDelegate, UIS
               let e2eEncrypted = userInfo["e2eEncrypted"] as? Bool
         else { return }
 
+        print("\(progressNumber.floatValue)")
```
mpivchev, line 547: Should remove this

mpivchev:
Looks good design wise! Only thing that might be a problem is if you have scrolled down and you try to upload a large file, the banner doesn't show up. Can we attach it to the top if you are scrolling?

mpivchev:
I still see this instead of an actual text

mpivchev:
Ah makes sense

mpivchev (approved):

PR #2555, opened 2023-08-02
Share add/update

TSI-amrutwaghmare:
This PR Contains fix related to share update.

Expected behaviour
New share get update and existing also keep in list
Actual behaviour
Existing shares get removed and only new share visible 
Steps to reproduce
create new 2 or more share 
App version
4.8.6

Current behaviour 
https://github.com/nextcloud/ios/assets/96108296/8b95ada3-0af6-4378-8936-41ea18032648

After Fix

https://github.com/nextcloud/ios/assets/96108296/8f8f0942-338e-42d6-8317-90a30ab97ce6

### DIFF
--- a/iOSClient/Data/NCManageDatabase+Share.swift
+++ b/iOSClient/Data/NCManageDatabase+Share.swift
@@ -81,8 +81,6 @@ extension NCManageDatabase {
         do {
             let realm = try Realm()
             try realm.write {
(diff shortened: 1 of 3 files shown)


mpivchev:
Looks good! Thanks for the fix. I will let @marinofaggiana approve, since he has better knowledge of Realm.

PR #2571, opened 2023-08-21
E2ee v20 Step - ONE

marinofaggiana:
- New E2ee V 2.0 step open ( more sicure with the share e2ee folder )
- Remain the compatibility with e2ee V 1.1 & 1.2
- Beta version in test

### DIFF
--- a/iOSClient/Data/NCManageDatabase+Metadata.swift
+++ b/iOSClient/Data/NCManageDatabase+Metadata.swift
@@ -235,7 +235,7 @@ extension tableMetadata {
         NCUtility.shared.isDirectoryE2EE(account: account, urlBase: urlBase, userId: userId, serverUrl: serverUrl)
     }
 
-    var isDirectoryTopE2EE: Bool {
+    var isDirectoryE2EETop: Bool {
         NCUtility.shared.isDirectoryE2EETop(account: account, serverUrl: serverUrl)
     }
 
@@ -249,7 +249,7 @@ extension tableMetadata {
         guard NCGlobal.shared.capabilityFileSharingApiEnabled else { return false }
 
         if NCGlobal.shared.capabilityE2EEApiVersion == NCGlobal.shared.e2eeVersionV20, isDirectoryE2EE {
-            return !isDirectoryTopE2EE
+            return !isDirectoryE2EETop
         } else if !e2eEncrypted && !isDirectoryE2EE {
             return true
         } else if NCGlobal.shared.capabilityServerVersionMajor >= NCGlobal.shared.nextcloudVersion26 && directory {
--- a/iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
+++ b/iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
@@ -1445,6 +1445,12 @@ extension NCCollectionViewCommon: UICollectionViewDataSource {
 
         guard let metadata = dataSource.cellForItemAt(indexPath: indexPath) else { return cell }
 
+        defer {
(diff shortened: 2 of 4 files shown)


iOSClient/Data/NCManageDatabase+Metadata.swift
```diff
@@ -232,20 +232,25 @@ extension tableMetadata {
     }
 
     @objc var isDirectoryE2EE: Bool {
-        NCUtility.shared.isDirectoryE2EE(serverUrl: serverUrl, account: account, urlBase: urlBase, userId: userId)
+        NCUtility.shared.isDirectoryE2EE(account: account, urlBase: urlBase, userId: userId, serverUrl: serverUrl)
+    }
+
+    var isDirectoryTopE2EE: Bool {
+        NCUtility.shared.isDirectoryE2EETop(account: account, serverUrl: serverUrl)
```
mpivchev, line 239: We have `isDirectoryTopE2EE` and `isDirectoryE2EETop` here, let's maybe choose a single way to name this

iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
```diff
@@ -1668,12 +1659,11 @@ extension NCCollectionViewCommon: UICollectionViewDataSource {
...
-        //
-
-        if !metadata.isSharable() {
+        // Disable Share Button
+        if appDelegate.disableSharesView || !metadata.isSharable() {
             cell.hideButtonShare(true)
         }
-        
+
         return cell
```
mpivchev, line 1667: If this should always be at the end, can't we put it in a `defer` block and move it to the top of this function?

iOSClient/Networking/E2EE/NCNetworkingE2EERename.swift
```diff
@@ -32,42 +32,38 @@ class NCNetworkingE2EERename: NSObject {
         return instance
     }()
 
+    let errorEncodeMetadata = NKError(errorCode: NCGlobal.shared.errorInternalError, errorDescription: NSLocalizedString("_e2e_error_encode_metadata_", comment: ""))
+    let fileAlreadyExists = NKError(errorCode: NCGlobal.shared.errorInternalError, errorDescription: "_file_already_exists_")
```
mpivchev, line 36: let's add `Error` at the end of these vars

iOSClient/Utility/NCUtility.swift
```diff
@@ -748,19 +748,47 @@ class NCUtility: NSObject {
...
+                if let directory = NCManageDatabase.shared.getTableDirectory(account: account, serverUrl: serverUrl) {
+                    if directory.e2eEncrypted {
+                        top = directory
+                    } else {
+                        return top
+                    }
+                } else { return top }
+            } else { return top }
+        }
+    }
```
mpivchev, line 790: This looks a bit complex with a lot of elses that do the same thing. Maybe something like this would be better:

```suggestion
func getDirectoryE2EETop(serverUrl: String, account: String) -> tableDirectory? {
    var serverUrl = serverUrl
    var top: tableDirectory? = nil
    
    while let url = URL(string: serverUrl)?.deletingLastPathComponent(),
          let directory = NCManageDatabase.shared.getTableDirectory(account: account, serverUrl: serverUrl) {
        
        if directory.e2eEncrypted {
            top = directory
        } else {
            return top
        }
        
        serverUrl = String(url.absoluteString.dropLast())
    }
    
    return top
}
```

mpivchev (changes requested):

mpivchev (approved):
Nice!

PR #2578, opened 2023-08-28
Fix #2577 : Added one second before `tipView` func call.

### LINKED ISSUE #2577: TipView's position is not correct for first app launch
### Steps to reproduce

1. Uninstall NextCloud iOS app (if already exists)
2. Install NextCloud iOS app
3. Login casually
4. On the home screen tipView's position isn't correct

### Expected behavior
The tipView should be at the center of the email

### Actual behavior
The tipView is at the top left corner.

When we launch the app, we call `showTip()` in `NCCollectionViewCommon's viewDidAppear`. This function is called even before data is fetched as a result of which frame / navigationData of navigationBarItem is not set properly giving an incorrect frame over which our tipView is shown.

However when we have received the data at least once, we are caching it, so in subsequent launches, we get the correct frame.

### Possible Solution:
Call the `showTip()` function only when data is fetched for the first app launch.

### Environment data

**iOS version:**  iOS: 16.1

**Nextcloud iOS app version:** see More > Settings

**Server operating system:**

**Web server:** Apache, nginx

**Database:**

**PHP version:**

**Nextcloud version:** see Nextcloud admin page

adityagi02:
As stated in #2577, it fixes the tipView as it should be at the centre of the user icon/email.

I've just added 1 second before `tipView` func call, so that after data is fetched which obviously takes less time than 1 second(mostly), will show the tip at exact location. 

https://github.com/nextcloud/ios/assets/77538183/2c5658c6-41cd-4b7f-a989-4f42e734523f

**I'm just a beginner, so please any suggestions will be helpful.**

### DIFF
--- a/iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
+++ b/iOSClient/Main/Collection Common/NCCollectionViewCommon.swift
@@ -223,8 +223,10 @@ class NCCollectionViewCommon: UIViewController, UIGestureRecognizerDelegate, UIS
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-
-        showTip()
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
+            self.showTip()
+        }
     }
 
     override func viewWillDisappear(_ animated: Bool) {


adityagi02:
> Thanks a lot ! remember for the next time to use DCO:
> 
> 
> 
> Git even has a -s | --signoff command line option to append this to your commit messages automatically.
> 
> 
> 
> 

Sure!, Is this a correct and acceptable solution? I know you'll be busy, but still if you could reply😃

adityagi02:
😬 DCO error, please approve one last time.

dogra25130:
Not very sure, but I feel adding 1 second delay might not fix the root cause. Although it looks much better, however, if we try to kill the app and open it again or rotate the app, then the tipView appears to be in the center. I feel we should keep the behavior consistent.



If we want to add the delay then we might have to add it of atleast 5 seconds till the time all the data has been fetched from the server to get tipView in the correct position.
 
However, this leads to another issue i.e. if the user's internet is slow then we might have to increase it even further, or in other words delay should be dynamic. So I feel instead of adding delay we should trigger `showTip()` function only when the entire data is fetched and loaded.

In this attached video I have just added a delay of 5 seconds.

https://github.com/nextcloud/ios/assets/32197474/99a17085-cbf9-4951-8c1c-763ecdf3eea9

Issue #2584, opened 2023-09-01
Interchange '+/-' with 'Select' from 'NCMedia+Menu'


adityagi02:
After #2583, we can interchange `+/-` with `Select` & `Cancel`, as it is a more frequently used option in the menu and will obviously reduce the hassle of selecting and then canceling selected media.

mpivchev:
That's a good idea. Let's see :)

Issue #2583, opened 2023-08-31
Add zoom in and zoom out to new media screen


mpivchev:
Once https://github.com/nextcloud/ios/issues/2399 is finished, we must add (preferably pinch to) zoom the photo gallery in and out like before

PR #2606, opened 2023-09-21
Media View in SwiftUI

### LINKED ISSUE #2399: Redo Media view with proper media aspect ratio
Show images with correct aspect ratio instead of square thumbnails.

mpivchev:
This PR adds a new media view written in SwiftUI

Things to be implemented:

- [x] https://github.com/nextcloud/ios/issues/2583
- [x]  https://github.com/nextcloud/ios/issues/2584
- [x] https://github.com/nextcloud/ios/issues/2619
- [x]  https://github.com/nextcloud/ios/issues/2613
- [x] #2612 
- [x] https://github.com/nextcloud/ios/issues/2672

image

This closes #2399


mpivchev:
@marinofaggiana What do you think about the performance when zoomed out fully?

IMO the performance is not good, and we can open a separate issue to try and fix it as much as possible by:

- Looking at the profiler and checking what causes low framerate
- Replacing normal UIImageView with [NukeUI](https://github.com/kean/Nuke) or Kingfisher. I already tried briefly using NukeUI, but it needs additional work

     -  Caching should be done by NukeUI instead of Realm
     -  Some SwiftUI code needs to be rewritten

mpivchev:
> ng time before all previous images are loaded (see old code).

Maybe this would also affect performance

mpivchev:
> When scroll the images down quickly the queue of preview is not deleted, the result is that you have to wait a very long time before all previous images are loaded (see old code).

Yeah maybe doing this would help with performance a bit

PR #2660, opened 2023-11-07
New media view - loading indicators

mpivchev:
This adds a loading indicator to the navigation bar:

It shows when:
- New or old media metadata is being loaded (thumbnail loading uses shimmering placeholders)
- Moving, copying, downloading or deleting media metadata

Additional things:
- Added Queuer to limit how many thumbnails are downloaded at a time
- Decoupled the main view model from the scroll view

### DIFF
--- a/iOSClient/Main/NCActionCenter.swift
+++ b/iOSClient/Main/NCActionCenter.swift
@@ -42,6 +42,8 @@ class NCActionCenter: NSObject, UIDocumentInteractionControllerDelegate, NCSelec
     let utilityFileSystem = NCUtilityFileSystem()
     let utility = NCUtility()
 
+    var onSelectionCancelled: (() -> Void)?
+
     // MARK: - Download
 
     @objc func downloadedFile(_ notification: NSNotification) {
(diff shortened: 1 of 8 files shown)


iOSClient/Main/NCActionCenter.swift
```diff
@@ -444,6 +446,7 @@ class NCActionCenter: NSObject, UIDocumentInteractionControllerDelegate, NCSelec
 
     // MARK: - Copy & Paste
 
+    @available(*, deprecated, message: "This calls the HUD view on dismiss, and the HUD will be removed throughout the app soon. Use copyToPasteboard(pasteboardOcIds:completion:) instead")
```
mpivchev, line 449: This is still used in other parts of the app, and for now should remain.

iOSClient/Main/NCActionCenter.swift
```diff
@@ -694,6 +734,109 @@ class NCActionCenter: NSObject, UIDocumentInteractionControllerDelegate, NCSelec
             appDelegate.window?.rootViewController?.present(navigationController, animated: true, completion: nil)
         }
     }
+
+    @available(*, deprecated, message: "This calls the HUD view on dismiss, and the HUD will be removed throughout the app soon. Use openSelectView(items:indexPath:didCancel:) instead")
```
mpivchev, line 738: This is still used in other parts of the app, and for now should remain.

Issue #2612, opened 2023-09-26
In Media View, show new photos as indicator instead of auto scrolling to top


mpivchev:
Related to: https://github.com/nextcloud/ios/pull/2606

Currently we scroll to the top when we load new photos in the Media View.

We should remember the position and instead add an indicator for new items, the way Twitter does it for example:

![image](https://github.com/nextcloud/ios/assets/6960329/dea5b514-c39a-485c-8f01-f131c52cbda2)

tobiasKaminsky:
Estimation: 3 days (8h)

mpivchev:
@nextcloud/designers @tobiasKaminsky @marinofaggiana 

This is how the button looks like. Is it ok?

<img width="555" alt="image" src="https://github.com/nextcloud/ios/assets/6960329/fe3ffc7b-d36b-4156-a1f8-ab6a20720cd0">

nimishavijay:
Looks super nice! Would only add like 2-4px padding on the right, other than that all ok 👍

PR #2674, opened 2023-11-20
New Media View - New Media indicator

mpivchev:
This adds a indicator when new media is loaded on top:

Also (mostly) fixes some bugs with:

- Jittery scroll up
- Scroll View not keeping its current position when new media is loaded


mpivchev:
This closes #2612

PR #2683, opened 2023-11-29
Livephoto Server NC 28

marinofaggiana:
Nextcloud 28 will be natively compatible with LivePhoto.
- livephoto detection structure changed (NextcloudKit included)
- new api to set up an uploaded file as LivePhoto
- start of a strategy to convert existing livephotos (at test phase only probable in background ... )

### DIFF
--- a/iOSClient/Data/NCManageDatabase+Metadata.swift
+++ b/iOSClient/Data/NCManageDatabase+Metadata.swift
@@ -249,7 +249,7 @@ extension tableMetadata {
         NCUtilityFileSystem().isDirectoryE2EETop(account: account, serverUrl: serverUrl)
     }
 
-    var livePhoto: Bool {
+    var isLivePhoto: Bool {
         !livePhotoFile.isEmpty
     }
 
@@ -1027,7 +1027,7 @@ extension NCManageDatabase {
 
     func getMetadataLivePhoto(metadata: tableMetadata) -> tableMetadata? {
 
-        guard metadata.livePhoto else { return nil }
+        guard metadata.isLivePhoto else { return nil }
 
         do {
             let realm = try Realm()
--- a/iOSClient/Networking/NCNetworking.swift
+++ b/iOSClient/Networking/NCNetworking.swift
@@ -691,7 +691,7 @@ class NCNetworking: NSObject, NKCommonDelegate {
                 utilityFileSystem.removeFile(atPath: utilityFileSystem.getDirectoryProviderStorageOcId(ocIdTemp))
             }
 
-            setLivephotoUpload(metadata: metadata)
+            uploadLivePhoto(metadata: metadata)
 
             NextcloudKit.shared.nkCommonInstance.writeLog("[SUCCESS] Upload complete " + serverUrl + "/" + fileName + ", result: success(\(size) bytes)")
             NotificationCenter.default.postOnMainThread(name: NCGlobal.shared.notificationCenterUploadedFile, userInfo: ["ocId": metadata.ocId, "serverUrl": metadata.serverUrl, "account": metadata.account, "fileName": metadata.fileName, "ocIdTemp": ocIdTemp, "error": error])
@@ -792,10 +792,10 @@ class NCNetworking: NSObject, NKCommonDelegate {
 
     // MARK: - Live Photo
 
-    func setLivephotoUpload(metadata: tableMetadata) {
+    func uploadLivePhoto(metadata: tableMetadata) {
 
         guard NCGlobal.shared.capabilityServerVersionMajor >= NCGlobal.shared.nextcloudVersion28,
-              metadata.livePhoto,
+              metadata.isLivePhoto,
               let metadata1 = NCManageDatabase.shared.getMetadata(predicate: NSPredicate(format: "account == %@ AND urlBase == %@ AND path == %@ AND fileName == %@ AND status == %d", metadata.account, metadata.urlBase, metadata.path, metadata.livePhotoFile, NCGlobal.shared.metadataStatusNormal)) else {
             return
         }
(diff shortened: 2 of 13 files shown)


iOSClient/AppDelegate.swift
```diff
@@ -285,6 +285,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         NCNetworking.shared.cancelDownloadTasks()
         presentPasscode { }
 
+        //TODO: TESTTTT
```
mpivchev, line 288: This needs to go

iOSClient/Data/NCManageDatabase+Metadata.swift
```diff
@@ -249,6 +249,10 @@ extension tableMetadata {
         NCUtilityFileSystem().isDirectoryE2EETop(account: account, serverUrl: serverUrl)
     }
 
+    var livePhoto: Bool {
```
mpivchev, line 252: probably better to rename it to `isLivePhoto`

iOSClient/Networking/NCNetworking.swift
```diff
@@ -790,6 +790,48 @@ class NCNetworking: NSObject, NKCommonDelegate {
         })
     }
 
+    // MARK: - Live Photo
+
+    func setLivephotoUpload(metadata: tableMetadata) {
```
mpivchev, line 795: Better name: `uploadLivePhoto`

iOSClient/Data/NCManageDatabase+Metadata.swift
```diff
@@ -734,6 +715,19 @@ extension NCManageDatabase {
         }
     }
 
+    func setMetadataLivePhotoServer(account: String, ocId: String) {
+
+        do {
+            let realm = try Realm()
+            try realm.write {
+                let result = realm.objects(tableMetadata.self).filter("account == %@ AND ocId == %@", account, ocId).first
+                result?.livePhotoServer = true
```
mpivchev, line 724: How is `livePhotoServer` different from `livePhotoFile` and `livePhoto`?

mpivchev (changes requested):

iOSClient/AppDelegate.swift
mpivchev: I don't think this was resolved

iOSClient/Data/NCManageDatabase+Metadata.swift
mpivchev: We should probably add this as docs to the fields. Give me some time

iOSClient/Data/NCManageDatabase+Metadata.swift
```diff
@@ -61,8 +61,8 @@ class tableMetadata: Object, NCUserBaseUrl {
     @objc dynamic var iconName = ""
     @objc dynamic var iconUrl = ""
     @objc dynamic var isExtractFile: Bool = false
-    @objc dynamic var livePhoto: Bool = false
     @objc dynamic var livePhotoFile = ""
+    @objc dynamic var livePhotoServer: Bool = false
```
mpivchev, line 65:
```suggestion
    /// A flag indicating if the file is sent as a live photo from the server, or if we should detect it as such and convert it client-side. 
    /// Usually this will be false if it's an old media and there is no `<metadata-files-live-photo xmlns="[http://nextcloud.org/ns\"/>]` flag in the metadata, and true if there is.
    @objc dynamic var isFlaggedAsLivePhotoByServer: Bool = false
```

iOSClient/Data/NCManageDatabase+Metadata.swift
```diff
@@ -61,8 +61,8 @@ class tableMetadata: Object, NCUserBaseUrl {
     @objc dynamic var iconName = ""
     @objc dynamic var iconUrl = ""
     @objc dynamic var isExtractFile: Bool = false
-    @objc dynamic var livePhoto: Bool = false
     @objc dynamic var livePhotoFile = ""
```
mpivchev, line 64:
```suggestion
    ///If this is not empty, the media is a live photo. New media gets this straight from server, but old media needs to be detected as live photo and converted client side. 
    @objc dynamic var livePhotoFilePath = ""
```

mpivchev (changes requested):

iOSClient/Networking/NCNetworking.swift
```diff
@@ -790,6 +790,48 @@ class NCNetworking: NSObject, NKCommonDelegate {
...
+        let serverUrlfileNamePath = metadata.urlBase + metadata.path + metadata.livePhotoFile
+        let serverUrlfileNamePath1 = metadata1.urlBase + metadata1.path + metadata1.livePhotoFile
+
+        Task {
+            let results = await NextcloudKit.shared.setLivephoto(serverUrlfileNamePath: serverUrlfileNamePath, livePhotoFile: metadata1.livePhotoFile)
+            print("Send LivePhoto metadata error \(results.error.errorCode)")
+
+            let results1 = await NextcloudKit.shared.setLivephoto(serverUrlfileNamePath: serverUrlfileNamePath1, livePhotoFile: metadata.livePhotoFile)
+            print("Send LivePhoto metadata1 error \(results1.error.errorCode)")
+        }
```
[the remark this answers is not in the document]
mpivchev: As we discussed, try to find better names for these.

iOSClient/Data/NCManageDatabase+Metadata.swift
mpivchev: Oops, will modify

mpivchev:
> > Should rename vars AND add the comments in both NextcloudKit and iOS app.
> 
> 
> 
> suggest .. remember need modify NextcloudKit, iOS app, database version ... and oldSchemaVersion  conversion

You can do this in new PR

PR #2766, opened 2024-01-31
Change Action sheet

mpivchev:
What is changed:

- Copy, Print, Save to photo gallery removed -> Now use Share
- Move or copy renamed to Move
- Set as available offline and remove as offline renamed to -> Download / Remove locally
- Changed top and bottom item margin from 8 to 6 to reduce space
- Changed some icons
- Title is now bold
- Delete action is now marked destructive
- Moved separator to after file name

Before: 

image

After:

image

Before:

image

After:

image


mpivchev:
- Added Download live photo option to live photos
- Change Move to Move or Copy
- Changed some more icons: details, rename, move, save as scan

image

- Changed title line break to truncate middle, to match files list line break:

    image

PR #2771, opened 2024-02-05
Media tab bar select

marinofaggiana:
- fix

### DIFF
--- a/iOSClient/Media/NCMediaTabbarSelect.swift
+++ b/iOSClient/Media/NCMediaTabbarSelect.swift
@@ -76,10 +76,11 @@ class NCMediaTabbarSelect: ObservableObject {
 
 struct MediaTabBarSelectView: View {
     @ObservedObject var tabBarSelect: NCMediaTabbarSelect
+    @Environment(\.verticalSizeClass) var sizeClass
 
     var body: some View {
         VStack {
-            Spacer().frame(height: 10)
+            Spacer().frame(height: sizeClass == .compact ? 5 : 10)
             HStack {
                 Spacer().frame(maxWidth: .infinity)
                 Group {
@@ -97,6 +98,7 @@ struct MediaTabBarSelectView: View {
                     tabBarSelect.delegate?.delete()
                 } label: {
                     Image(systemName: "trash")
+                    .imageScale(sizeClass == .compact ? .medium : .large)
                 }
                 .tint(.red)
                 .disabled(tabBarSelect.selectCount == 0)
--- a/Nextcloud.xcodeproj/project.pbxproj
+++ b/Nextcloud.xcodeproj/project.pbxproj
@@ -5308,8 +5308,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/nextcloud/NextcloudKit";
 			requirement = {
-				kind = exactVersion;
-				version = 2.9.5;
+				branch = develop;
+				kind = branch;
 			};
 		};
 		F788ECC5263AAAF900ADC67F /* XCRemoteSwiftPackageReference "MarkdownKit" */ = {


iOSClient/Media/NCMediaCommandView.swift
```diff
@@ -28,7 +28,8 @@ class NCMediaCommandView: UIView {
 
     @IBOutlet weak var title: UILabel!
     @IBOutlet weak var activityIndicator: UIActivityIndicatorView!
-    @IBOutlet weak var selectButton: UIButton!
+    @IBOutlet weak var selectcancelButton: UIButton!
```
mpivchev, line 31: Should be selectCancel

iOSClient/Media/NCMediaCommandView.swift
```diff
@@ -67,12 +68,60 @@ class NCMediaCommandView: UIView {
         gradient.frame = bounds
     }
 
-    @IBAction func selectButtonPressed(_ sender: UIButton) {
```
mpivchev, line 70: Same here

iOSClient/Media/NCMediaCommandView.swift
```diff
@@ -67,12 +68,60 @@ class NCMediaCommandView: UIView {
...
+    @IBAction func selectcancelButtonPressed(_ sender: UIButton) {
+
+        mediaView.isEditMode = !mediaView.isEditMode
+        setSelectcancelButton()
+    }
+
+    func setSelectcancelButton() {
```
mpivchev, line 77: Here too

iOSClient/Media/NCMediaCommandView.swift
```diff
@@ -67,12 +68,60 @@ class NCMediaCommandView: UIView {
...
+            selectcancelButtonTrailing.constant = 46
+            mediaView.tabBarSelect?.hide(animation: true)
+        }
 
-        mediaView.isEditMode = true
         mediaView.collectionView.reloadData()
+    }
+
+    func setTitleDate() {
+
```
mpivchev, line 96: What's the point of this empty line here? I see it on so many files. Can you change your settings?

iOSClient/Media/NCMediaTabbarSelect.swift
```diff
@@ -25,8 +25,7 @@ import UIKit
 import SwiftUI
 
 protocol NCTabBarSelectDelegate: AnyObject {
```
mpivchev, line 27: Please rename it to NCMediaTabBarSelect + the delegate. There are multiple tab bars now

mpivchev (changes requested):
Will check on simulator now

iOSClient/Media/NCMediaCommandView.swift
mpivchev: sure, but they are still separate words, so camel case is needed. maybe `selectOrCancel` would work

iOSClient/Media/NCMediaCommandView.swift
mpivchev: New functions made by you always have an empty line on the top. Can you change this?

mpivchev:
More things:
- Replace Cancel with Done to match other screens and tabs
- Top bar should animate when changing color (check SwiftUI version)
- Add .thinMaterial blur to top bar buttons (check SwiftUI version)
- When pressing delete on media, change dialog to action sheet (to match the rest)
    - Also a simple "Delete selected media? -Yes - Cancel" would be enough. Too much text IMO.
    
    image

- Zoom should not be a collapsed option, IMO it's annoying to always tap twice, especially if you wanna zoom multiple times

    image

- Change "Select the Media folder" to "Set Media folder" to match Android. Also reads better.

- Gradient is too harsh, you can see the line here. Make it smoother (check iOS Photos gradient)
- 
   image

mpivchev:
Some things still missing:


-Top bar should animate when changing color (check SwiftUI version)

- When pressing delete on media, change dialog to action sheet (to match the rest)
    - Also a simple "Delete selected media? -Yes - Cancel" would be enough. Too much text IMO.
    
    image

- Zoom should not be a collapsed option, IMO it's annoying to always tap twice, especially if you wanna zoom multiple times

    image

     - How it should look:
     
        image


- Gradient is too harsh, you can see the line here. Make it smoother (check iOS Photos gradient)
- 
   image

- View options are not marked as selected, no way of knowing what is chosen. 


Can add a checkmark via `UIAction(state: bool ? .on : .off)`

mpivchev:
Besides this tab bar looks nice :)

PR #2783, opened 2024-02-08
Skip creation screen on creating new text file

mpivchev:
This removes the creation screen when pressing "Create new document" and instead saves the file as Untitled.txt in the current directory.

Also added loading indicators when loading NextcloudText or Collabora files.


mpivchev:
Create text document (Nextcloud text) -> Changed extension to .md
Create new document (Collabora) -> Code reverted to original. Collabora actions are not part of this ticket.

PR #2814, opened 2024-02-27
Add swipe actions to List View files and folders

mpivchev:
This adds swipe actions.

Note: Originally we were supposed to have Share and Delete as actions, however Delete is very complex in that it can either be Delete or Leaving Share, and requires a confirmation dialog. IMO the swipe options should be for more simple actions that don't require additional confirmation and so on.

Simulator Screen Recording - iPhone 15 Pro - 2024-02-27 at 10 40 30


mpivchev:
@marinofaggiana I added swipe to delete action, the only issue is that the deletion takes a long time.

PR #2855, opened 2024-03-19
Tab bar refactoring

marinofaggiana:
- Rewrite the manage Tab bar (default/select)  on ConnectionViewCommon + Trash
- Fixed some bug in Trash
- Code cleaning
- Normalized log [INFO], [ERROR]

- [x]  Verify

### DIFF
--- a/Nextcloud.xcodeproj/project.pbxproj
+++ b/Nextcloud.xcodeproj/project.pbxproj
@@ -84,8 +84,6 @@
 		F31F69612A2F907800162F76 /* __Snapshots__ in Resources */ = {isa = PBXBuildFile; fileRef = F31F69602A2F907800162F76 /* __Snapshots__ */; };
 		F31F69642A2F929600162F76 /* PreviewSnapshots in Frameworks */ = {isa = PBXBuildFile; productRef = F31F69632A2F929600162F76 /* PreviewSnapshots */; };
 		F321DA8A2B71205A00DDA0E6 /* NCTrashSelectTabBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = F321DA892B71205A00DDA0E6 /* NCTrashSelectTabBar.swift */; };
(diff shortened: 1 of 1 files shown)


mpivchev:
https://github.com/nextcloud/ios/assets/6960329/bdad787c-22d5-4f7f-aa48-56414ced0ea9

Bottom bar disappears too late, maybe should be in `willDisappear` instead of `didDisappear`

mpivchev:
Uploading RPReplay_Final1711027815.MP4…

Cancel button stays on when switching accounts

mpivchev:
image

Top bar on landscape has no margins

iOSClient/Trash/Cell/NCTrashGridCell.swift
```diff
@@ -0,0 +1,287 @@
...
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+import UIKit
+
+class NCTrashGridCell: UICollectionViewCell, UIGestureRecognizerDelegate, NCCellProtocol, NCTrashCellProtocol {
```
mpivchev, line 26: Can we remove NCCellProtocol, NCTrashCellProtocol? I don't think we need them at this point

iOSClient/Trash/Cell/NCTrashGridCell.swift
```diff
@@ -0,0 +1,287 @@
...
+        set { imageStatus = newValue }
+    }
+    var fileLocalImage: UIImageView? {
+        get { return imageLocal }
+        set { imageLocal = newValue }
+    }
+    var fileFavoriteImage: UIImageView? {
+        get { return imageFavorite }
+        set { imageFavorite = newValue }
+    }
```
mpivchev, line 85: This is a lot of boilerplate that is useless IMO, it's hard to remove as it's in many parts of the code, but good to remove it nontheless

mpivchev (approved):
A lot of code, but looks ok on first glance. Still bugs to fix and other comments.

PR #2898, opened 2024-04-25
Fix light and dark color sensitivity

mpivchev:
There were cases where some colors would not be considered too light or dark, leading to this:

image

Adjusted values to catch those.

Now:

image


mpivchev:
@marinofaggiana I don't think it will have any negative impacts

PR #2908, opened 2024-04-30
Enable Login v2

mpivchev:
This enables Login v2 and adds support for TLS client certificates.

@marinofaggiana  We can remove all v1 code with removal of oldest supported NC version


Brand/NCBrand.swift
```diff
@@ -68,7 +68,7 @@ let userAgent: String = {
     @objc public var use_themingColor: Bool = true
     @objc public var use_themingLogo: Bool = false
     @objc public var use_storeLocalAutoUploadAll: Bool = false
-    @objc public var use_loginflowv2: Bool = false                                              // Don't touch me !!
+    @objc public var use_loginflowv2: Bool = true                                              // Don't touch me !!
```
mpivchev, line 71: @marinofaggiana The plan is to use v2 by default from now on, and v1 will be obsolete. Maybe we can remove this? Not sure if the brander will break

mpivchev: Is it ok to remove it?

mpivchev:
Added the option to do the flow via external browser @tobiasKaminsky

iOSClient/Login/NCLoginWeb.swift
```diff
@@ -47,6 +47,9 @@ class NCLoginWeb: UIViewController {
     var loginFlowV2Endpoint = ""
     var loginFlowV2Login = ""
 
+    // Opens the login URL in external browser instead of in app. User must manually go back to the app.
+    let loginFlowv2ExternalBrowser = false
```
mpivchev, line 51: @marinofaggiana Please test with this enabled and disabled

iOSClient/Login/NCLoginWeb.swift
```diff
@@ -169,21 +174,32 @@ class NCLoginWeb: UIViewController {
...
+            webView.load(request)
+        }
+    }
 
-        webView.load(request)
+    private func pollLogin() {
+        NextcloudKit.shared.getLoginFlowV2Poll(token: self.loginFlowV2Token, endpoint: self.loginFlowV2Endpoint) { server, loginName, appPassword, _, error in
+            if error == .success, let server, let loginName, let appPassword {
+                self.createAccount(server: server, username: loginName, password: appPassword)
+            }
```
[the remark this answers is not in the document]
mpivchev: I stopped using `NCLoginWeb` for v2 and this screen stopped appearing. Now we only use `NCLogin`. `NCLoginWeb` can be renamed to `NCLoginProviders` as now only external providers use it.

mpivchev:
@marinofaggiana Please change `openLogin` function in `AppDelegate`, as that still calls `NCLoginWeb`

Issue #2130, opened 2022-08-21
Nextcloud File Integration not working at all


solick:
### Steps to reproduce

1. Open iOS Files app
2.Click on Nextcloud to open next cloud folder overview

### Expected behaviour
Root Directory view

### Actual behaviour

After a short load animation, 0 files are shown

### Environment data

**iOS version:** iOS 14.6.1

**Nextcloud iOS app version:** 4.4.2.2

**Server operating system:** Ubuntu

**Web server:** nginx

**Database:** MySQL

**PHP version:*7.4

**Nextcloud version:** 24.0.4

JDDoyon:
Same issue for me...

Environment data

Nextcloud iOS app version: 4.4.2.0

Server operating system: Ubuntu

Web server: nginx

Database: MySQL

PHP version: 7.3.31-1

Nextcloud version: 23.0.8

Supergamerx3000:
Open iOS Files app
2.Click on Nextcloud to open next cloud folder overview
Expected behaviour

Root Directory view

Actual behaviour

After a short load animation,

content not available

the content could not be displayed due to an unknown error

Environment data

iOS version: iOS 15.7

Nextcloud iOS app version: 4.4.4.0

Server operating system: Ubuntu

Web server: nginx

Database: MySQL

**PHP version:*7.4

Nextcloud version: 24.0.5

Only one User is affected by this error

solick:
Update: With last update I can now access next cloud via Favorites but Nextcloud itself is not shown as Place.

andyxheli:
Same issue

bronson:
I have the same issue except, in the Nextcloud folder in the files app, I see one file named “communication” of 128KB in size. It appears to be the mobile app’s logfile.

Is there any way to make the files integration work?

solick:
more than 9 month open and still this issue is not fixed with the current versions. Any information on that from the developers would be highly appreciated!

gitstn:
I am having same issue. I only see Communications file in the Nextcloud folder in the files app. I have tried deleting the nextcloud app and reinstalling it but no go. Would appreciate a response.

saghul:
Even if Files integration is enabled in the app I had to turn it on in the Files app too. While on the locations list, tap the 3 dots, then edit, and turn it on. That worked for me.

![image](https://user-images.githubusercontent.com/317464/236633482-224e805d-f1d8-4df5-94db-baab5fad6080.png)

gitstn:
[...]

PR #2911, opened 2024-05-04
Give hint for non-appearing Nextcloud in Files Locations on initial install

itsthejb:
Hi everyone,

**Note this is incomplete since the proposed hint solution would require localization in all languages.** However this seems like a low-cost way to potential work around an iOS-imposed problem.

# Problem

Unfortunately when installing Nextcloud for the first time on a device, it will not appear in the Files app Locations list until manually added to the list through the app's more icon + Edit:

IMG_31385ED7A320-1

I've been using the app happily for many years, however have been repeatedly tripped up by this after each device upgrade. From web search, it appears many people are misled by this problem:

- https://help.nextcloud.com/t/ios-nextcloud-app-and-files-manager-integration-no-longer-works/136314
- https://help.nextcloud.com/t/solved-it-seems-nextcloud-ios-5-0-1-0-app-integration-not-working/181973
- https://github.com/nextcloud/ios/issues/2130
- https://help.nextcloud.com/t/ios-integration-does-not-work/140519

So it seems to me that something should be done to at least help hint users in the right direction.

# Proposed solution

This seems like something which could best be handled through onboarding. However the app doesn't have any right now. Instead, since there is already the "Disable Files App integration" button in Advanced setting with a hint footer text, perhaps this could at least be leveraged to hint the user into realising the problem.

So, I propose slightly extending the text here in the hope it leads users in the right direction

### DIFF
--- a/iOSClient/Supporting Files/en.lproj/Localizable.strings
+++ b/iOSClient/Supporting Files/en.lproj/Localizable.strings
@@ -371,7 +371,7 @@
 "_advanced_"                    = "Advanced";
 "_permissions_"                 = "Permissions";
 "_disable_files_app_"           = "Disable Files App integration";
-"_disable_files_app_footer_"    = "Do not permit the access of files via the iOS Files application";
+"_disable_files_app_footer_"    = "Do not permit the access of files via the iOS Files application. Note that on initial install Nextcloud will not appear in Files' Locations list until manually added via More > Edit";
 "_trial_"                       = "Trial";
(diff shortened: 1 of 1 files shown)


itsthejb:
Hi @marinofaggiana,

Thanks for your response. I'm referring to this text:

IMG_14EC1A7A80E2-1

I searched the string in the code, so I think this is the one. The proposal is just to add a note here about the unexpected/confusing OS behaviour

PR #2918, opened 2024-05-11
Openssl new version 3.1.5004

DavidIsermann:
Changes OpenSSL Version to 3.1.5004 to include apples privacy manifest which is now needed to upload the app to the App Store.


asadamatic:
I'm facing the same issue on App Store.

pultho:
@marinofaggiana could you please specify more precisely why you just closed this (and other) MR. I'm having difficulties understanding your problem here.

asadamatic:
@marinofaggiana For new app the privacy manifest is also needed for third party, privacy-impacting SDKs.
Though, it's understandable that there can be compatibility issues with the version upgrade, but can you give a timeline for this upgrade?

PR #2966, opened 2024-07-01
New version 5.4.0

marinofaggiana:
- A new account setting view (available from Files top-left menu).
- Scan document: it is possible to modify a document after scanning it.
- Login Version 2 has been implemented (authentication now through external browser)
- TLS support (Increases security by using a mandatory client/server-side certificate)
- The settings section has been completely rewritten, with the addition of a Display section 
- Change filename mask (now present only in Settings/Advanced): by default, when a file is uploaded, it automatically gets the following format: yy-mm-dd hh-mm-ss plus a 4-digit counter. If you do not like this format, you can change it except for the final 4-digit counter, which cannot be omitted.
- Many other improvements to the code to make it more stable and optimized


mpivchev (approved):

Wangtt185 (commented):
q

PR #2967, opened 2024-07-01
Share screen redesign

mpivchev:


mpivchev:
> @mpivchev  Can you remove (replace) the DropDown lib ?

Not possible

PR #3021, opened 2024-08-12
Fix stuck re-login

mpivchev:
Fix being stuck on the log in screen when you are logged out


mpivchev:
@marinofaggiana If I make something ready for review you can always test it

PR #3042, opened 2024-09-05
"Enforce servers" brand option

mpivchev:


mpivchev:
Also fixed QR code icon being low quality

image

Issue #2615, opened 2023-09-30
Auto Upload Feature Request: Sync favorites with Apple Photos & Sync only favorited photos


ppnplus:
I have two feature requests.

First is a feature to sync favorites with Apple photo library, where favoriting photos in either app would mark it as favorites in either app. This feature is derived from [Google Photos' existing feature](https://support.google.com/photos/answer/9845656?hl=en). This can be helpful, as I sometimes favorite photos for my easy access, including slides.

Lastly, a feature to only sync favorites in the Apple photo library, which was requested by a [Nextcloud forums member](https://help.nextcloud.com/t/ios-sync-only-photos-that-are-marked-as-favorites/134920). In that user's case, it can be useful if taking multiple photos and wanting only the desired (favorited) content to be uploaded.

cclarke591:
You may want to format this into 2 separate "issues" or feature requests. Additionally I would also title it as [Feature Request] or [enhancement] based on existing labels.

Mi-Sky:
+1 vote for a feature to only sync favorites in the Apple photo library. I would also like to offer a bounty for it, how?

osovitskiy:
Hi devs,
I see that the feature was implemented in #3105, and then the original implementation was superseded by #3360. When I'm trying to set up favourite-only autoupload on iOS 18 device, I don't see any option to choose "Favourite" album, the only albums avaialble are "Camera roll" and "Videos":

![Image](https://github.com/user-attachments/assets/c67d5b99-e063-4987-8faf-2553446517db)

iOS 18 & NextCloud client version 6.6.0.4

Could someone please clarify the status of the favourite-only autoupload feature for iOS client?

PR #3102, opened 2024-10-10
add option to sync favorites only.

sp00ktober:
Implements #2615 

This adds an option to the Auto Upload settings to only sync those assets marked as favorites through the Apple Photos App.

I did some testing in the simulator and with my own Nextcloud server. I am open for improvements of this PR, especially regarding localization. Not sure if i did this the right way or not.

This is just a super handy feature for myself and i was missing it in the app, so i added it.


sp00ktober:
I am closing this one and create another one to the develop branch according to the readme. Sorry for the trouble.

PR #3183, opened 2024-11-11
Unified Search Fallback Presentation

i2h3:
After some back and forth because of the limited on iOS to programmatically open contacts and calendar events, here is a simple change which enables the fallback for unified search result items to be opened in the server's web user interface if every other code path before fails.

Without this, taps on contact or calendar event items do not cause anything to happen.


iOSClient/Main/Collection Common/NCCollectionViewCommon+CollectionViewDelegate.swift
```diff
@@ -82,6 +84,10 @@ extension NCCollectionViewCommon: UICollectionViewDelegate {
                                                                                                sceneIdentifier: self.controller?.sceneIdentifier) {
 
                 NCNetworking.shared.download(metadata: metadata, withNotificationProgressTask: true)
+
+                if metadata.url.isEmpty == false {
```
mpivchev, line 88:
```suggestion
                if !metadata.url.isEmpty {
```

i2h3: Fine with me. I usually prefer the more explicit expression because it cannot be overlooked as easily as the short form.

iOSClient/Main/Collection Common/NCCollectionViewCommon+CollectionViewDelegate.swift
```diff
@@ -82,6 +84,10 @@ extension NCCollectionViewCommon: UICollectionViewDelegate {
                                                                                                sceneIdentifier: self.controller?.sceneIdentifier) {
 
                 NCNetworking.shared.download(metadata: metadata, withNotificationProgressTask: true)
+
+                if !metadata.url.isEmpty {
+                    NCViewer().view(viewController: self, metadata: metadata, image: nil)
+                }
```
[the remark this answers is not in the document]
i2h3: I misread the existing code after back and forth in changes. I pushed an updated branch which distinguishes between files to download and everything else at that point.

PR #3204, opened 2024-11-22
Close #3184: Download Limit Support

### LINKED ISSUE #3184: Download Limit Support
See [parent issue in meta repository for files clients](https://github.com/nextcloud/files-clients/issues/54).

**Important consideration**: [The optional server-side app](https://github.com/nextcloud/files_downloadlimit) for limiting downloads can be disabled or not even installed.

Add UI tests, if feasible.

i2h3:
- Added tableDownloadLimit entity to app database.
- Extended capability query to also consider download limit app.
- Extended capabilities list view for display of download limit availability.
- Extended share detail user interface to mimic the web user interface for managing download limits.
- Every time WebDAV properties of a file are retrieved, its associated download limits are removed and recreated.

Housekeeping:
- Outsourced NCShareDateCell into dedicated source code file.
- Outsourced NCShareToggleCell into dedicated source code file.

Notes:
- In my first attempt I had a detail view in the download limit row of the advanced share options showing the remaining number of downloads. However, that required to inject and retain the download limit entity object into the complicated share table configuration object. That, in turn, results in inconsistent data state due to invalid and outdated references. To resolve those issues, the assembly of the advanced share options user interface needs some refactoring which appears to expansive at this point and I prefer to leave it as it is for now.

### DIFF
--- a/Brand/Database.swift
+++ b/Brand/Database.swift
@@ -26,4 +26,4 @@ import Foundation
 // Database Realm
 //
 let databaseName                    = "nextcloud.realm"
-let databaseSchemaVersion: UInt64   = 367
+let databaseSchemaVersion: UInt64   = 368
--- a/Nextcloud.xcodeproj/project.pbxproj
+++ b/Nextcloud.xcodeproj/project.pbxproj
@@ -5913,8 +5913,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/nextcloud/NextcloudKit";
 			requirement = {
-				kind = exactVersion;
-				version = 5.0.2;
+				branch = files_downloadlimit;
+				kind = branch;
 			};
 		};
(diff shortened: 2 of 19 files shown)


i2h3:
This depends on [the still pending merge request for required changes in NextcloudKit](https://github.com/nextcloud/NextcloudKit/pull/107).

mpivchev (commented):
image

I see some whitespace warnings, can you enable this if you haven't already?

mpivchev:
Can you add an On Off label on the right side of "Limit Downloads" ?

Example:
image

i2h3:
@mpivchev Now I remember what the problem was. It worked only partially due to the complicated `UITableView` setup. I had an implementation where the initial state was presented but due to the static retention of `NCShareConfig`, how it is passed through and the metadata and how cells are constructed it appeared to me that I can only get this through with extended refactoring which I considered out of scope.

mpivchev (approved):

PR #3218, opened 2024-12-02
Updated Localizations

i2h3:
As discussed with @tobiasKaminsky, I checked [the availability of localizations on Transifex](https://app.transifex.com/nextcloud/nextcloud/ios/) and reinstated some localizations which have at least 50% reviewed translations like it is handled on Android. There were leftovers from the following localizations in the project which I reinstated:

* Croatian
* Lao
* Slovenian

I am new to the localization process. Is an update of the `iOSClient/.tx/config` file necessary?


i2h3:
> why you have added the Lang here ?

I did not notice that the files are registered there in the project tree. This is the conventional way Xcode does it by itself when new localizations are added. Where else do they belong? I do not see the others. The storyboard is localizable. I do not exactly understand the problem here.

PR #3223, opened 2024-12-05
#3169 fix ssl client certificate

zero0cool0:
Issue #3169

I tested SSL Client Certificate authentication and found that it didn't really work for me.

`NCLogin` does not set itself as `certificateDelegate` of `NCNetworking` hence its delegate methods are never called if NSURLAuthenticationMethodClientCertificate is received in `NCNetworking.authenticationChallenge didReceive challenge`

UIAlertControllers in `NCLogin` are not presented on the main thread

These two changes correctly prompt for a certificate file to pick and also ask for a certificate password.

However, once the login flow is restarted after `p12Data` is set, `authenticationChallenge didReceive challenge` wasn't called once more, I suspect due to a reuse of the underlying SSL session.

I found that instead of the default action, it is better to cancel the authentication challenge if `didAskForClientCertificate()` is called on the delegate. This leads to the challenge being received and now properly answered with the client certificate.

### DIFF
--- a/iOSClient/Login/NCLogin.swift
+++ b/iOSClient/Login/NCLogin.swift
@@ -188,6 +188,8 @@ class NCLogin: UIViewController, UITextFieldDelegate, NCLoginQRCodeDelegate {
                  return outgoing
              }
         }
+        
+        NCNetworking.shared.certificateDelegate = self
     }
 
     override func viewDidAppear(_ animated: Bool) {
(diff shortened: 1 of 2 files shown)


mpivchev:
Hi, this was tested and I can confirm the "Connection Lost" error is now fixed, and you can properly choose a certificate. Thanks :)

mpivchev (approved):

PR #3251, opened 2024-12-23
Add option for in-app browser log in

mpivchev:
V2 did not allow for in app browser log in, and this option is now added


Brand/NCBrand.swift
```diff
@@ -73,6 +73,7 @@ let userAgent: String = {
     var doNotAskPasscodeAtStartup: Bool = false
     var disable_source_code_in_settings: Bool = false
     var enforce_passcode_lock = false
+    var use_in_app_browser_for_login = true
```
mpivchev, line 76: @marinofaggiana This is needed for a patch, will revert later

iOSClient/Login/NCLogin.swift
```diff
@@ -491,4 +508,63 @@ extension NCLogin: ClientCertificateDelegate, UIDocumentPickerDelegate {
...
+                let controller = UIApplication.shared.firstWindow?.rootViewController as? NCMainTabBarController
+                NextcloudKit.shared.getLoginFlowV2Poll(token: loginFlowV2Token, endpoint: loginFlowV2Endpoint) { [self] server, loginName, appPassword, _, error in
+                    if error == .success, let urlBase = server, let user = loginName, let appPassword {
+                        loginFlowInProgress = true
+                        ncLoginPollModel.isLoading = true
+
+                        NCAccount().createAccount(urlBase: urlBase, user: user, password: appPassword, controller: controller) { account, error in
+
+                            if error == .success {
+                                let window = UIApplication.shared.firstWindow
```
i2h3, line 530: I quickly checked and realized that this sabotages the use of multiple windows on iPad OS. I have Nextcloud in split view with the Music app. Then I have another split view with Nextcloud Talk. When activating the first split view and selecting to add another account in the Nextcloud app there, the window gets destroyed, Music becomes full screen and the other Nextcloud window in the other split view shows the onboarding. Maybe it is not related directly to this line or merge request but it appears to be a general problem.

**Edit**: I mean **scenes**, not windows on iPad OS. The _n_ possible instances of a single app's user interface.

iOSClient/Login/Poll/NCLoginPoll.swift
```diff
@@ -0,0 +1,75 @@
...
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
```
i2h3, line 22: Wrong license header.

iOSClient/Login/Poll/NCLoginPoll.swift
```diff
@@ -0,0 +1,75 @@
...
+
+struct NCLoginPoll: View {
+    let loginFlowV2Login: String
+
+    @ObservedObject var model: NCLoginPollModel
+    @Environment(\.dismiss) private var dismiss
+
+    var body: some View {
+        VStack {
+            Text(NSLocalizedString("_poll_desc_", comment: ""))
```
i2h3, line 35: Why use `NSLocalizedString` here? SwiftUI initializers are designed for localized string and `Text("_poll_desc_")` should already do the trick. 🤔 

https://developer.apple.com/documentation/swiftui/preparing-views-for-localization

iOSClient/Login/Poll/NCLoginPollModel.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  NCLoginPollModel.swift
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 16.12.24.
+//  Copyright © 2024 Marino Faggiana. All rights reserved.
+//
```
i2h3, line 7: Wrong license header.

i2h3 (commented):
It has already been merged but maybe you can still consider my comments.

PR #3257, opened 2025-01-03
Auto upload from specific Albums

mpivchev:
[More info soon]


iOSClient/Extensions/PHAssetCollection+Extension.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  PHAssetCollection+Extension.swift
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 12.12.24.
+//  Copyright © 2024 Marino Faggiana. All rights reserved.
+//
```
i2h3, line 7:
```suggestion
// SPDX-FileCopyrightText: Nextcloud GmbH
// SPDX-FileCopyrightText: 2024 Milen Pivchev
// SPDX-License-Identifier: GPL-3.0-or-later
```

iOSClient/Settings/AutoUpload/Albums.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Albums.swift
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 12.12.24.
+//  Copyright © 2024 Marino Faggiana. All rights reserved.
+//
```
i2h3, line 7:
```suggestion
// SPDX-FileCopyrightText: Nextcloud GmbH
// SPDX-FileCopyrightText: 2024 Milen Pivchev
// SPDX-License-Identifier: GPL-3.0-or-later
```

iOSClient/Settings/SelectAlbum/AlbumModel.swift
```diff
@@ -0,0 +1,112 @@
+//
+//  AlbumModel.swift
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 12.12.24.
+//  Copyright © 2024 Marino Faggiana. All rights reserved.
+//
```
i2h3, line 7:
```suggestion
// SPDX-FileCopyrightText: Nextcloud GmbH
// SPDX-FileCopyrightText: 2024 Milen Pivchev
// SPDX-License-Identifier: GPL-3.0-or-later
```

iOSClient/Settings/SelectAlbum/SelectAlbumView.swift
```diff
@@ -0,0 +1,146 @@
+//
+//  SelectAlbumView.swift
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 20.11.24.
+//  Copyright © 2024 Marino Faggiana. All rights reserved.
+//
```
i2h3, line 7:
```suggestion
// SPDX-FileCopyrightText: Nextcloud GmbH
// SPDX-FileCopyrightText: 2024 Milen Pivchev
// SPDX-License-Identifier: GPL-3.0-or-later
```

iOSClient/Networking/NCAutoUpload.swift
```diff
@@ -235,29 +240,62 @@ class NCAutoUpload: NSObject {
...
-            if selector == NCGlobal.shared.selectorUploadAutoUpload,
-               let idAssets = self.database.getPhotoLibraryIdAsset(image: tableAccount.autoUploadImage, video: tableAccount.autoUploadVideo, account: account) {
-                assets.enumerateObjects { asset, _, _ in
-                    var creationDateString = ""
-                    if let creationDate = asset.creationDate {
-                        creationDateString = String(describing: creationDate)
+
+            if assetCollections.isEmpty {
+                let assetCollection = PHAssetCollection.fetchAssetCollections(with: .smartAlbum, subtype: PHAssetCollectionSubtype.smartAlbumUserLibrary, options: nil)
+                guard let assetCollection = assetCollection.firstObject else { return completion(nil) }
```
i2h3, line 246: Do I see it correctly that this is the only relevant difference between the large `if` and `else` blocks which mostly are identical? Looks like redundant code. The fetching in case of `assetCollections.isEmpty` probably can be reduced by many lines.

iOSClient/Settings/AutoUpload/NCAutoUploadView.swift
```diff
@@ -27,136 +27,171 @@ import UIKit
 
 /// A view that allows the user to configure the `auto upload settings for Nextcloud`
 struct NCAutoUploadView: View {
-    @ObservedObject var model: NCAutoUploadModel
+    @StateObject var model: NCAutoUploadModel
+    @StateObject var albumModel: AlbumModel
+    @State private var showUploadFolder: Bool = false
+    @State private var showSelectAlbums: Bool = false
```
i2h3, line 33: Inconsistent access control (some `private`, some `internal`).

iOSClient/Settings/NCKeychain.swift
```diff
@@ -609,6 +609,18 @@ import KeychainAccess
...
+
+    func setAutoUploadAlbumIds(account: String, albumIds: [String]) {
+        let key = "AlbumIds" + account
+        keychain[key] = albumIds.joined(separator: ",")
+    }
+
+    func getAutoUploadAlbumIds(account: String) -> [String]? {
+        let key = "AlbumIds" + account
+        return try? keychain.get(key)?.components(separatedBy: ",")
+    }
```
i2h3, line 622: Why the optional? Replace with an empty array returned in case of error instead, then you do not have to write the nil-coalescing (`?? []`) redundantly at the call points. The unused additional semantics of "something went wrong" in case of a `nil` return value are not used.
```suggestion
    func getAutoUploadAlbumIds(account: String) -> [String] {
        let key = "AlbumIds" + account
        return try? keychain.get(key)?.components(separatedBy: ",") ?? []
    }
```

iOSClient/Settings/AutoUpload/NCAutoUploadModel.swift
```diff
@@ -66,6 +65,10 @@ class NCAutoUploadModel: ObservableObject, ViewOnAppearHandling {
         NCSession.shared.getSession(controller: controller)
     }
 
+//    var autoUploadAlbumIds: [String] {
+//        NCKeychain().getAutoUploadAlbumIds(account: controller?.account ?? "") ?? []
+//    }
+
```
i2h3, line 71: Dead code which apparently can be removed.

iOSClient/Settings/AutoUpload/NCAutoUploadView.swift
```diff
@@ -27,136 +27,171 @@ import UIKit
...
 
     var body: some View {
-        Form {
-            /// Auto Upload
-            Section(content: {
-                Toggle(NSLocalizedString("_autoupload_", comment: ""), isOn: $model.autoUpload)
-                    .tint(Color(NCBrandColor.shared.getElement(account: model.session.account)))
-                    .onChange(of: model.autoUpload) { newValue in
-                        model.handleAutoUploadChange(newValue: newValue)
+        VStack {
```
i2h3, line 36: The `VStack` has the `Form` with its modifiers as its only child. What is it required for? I cannot tell. Looks removable to me.

iOSClient/Settings/AutoUpload/NCAutoUploadView.swift
```diff
@@ -27,136 +27,171 @@ import UIKit
...
-            .onChange(of: model.autoUploadSubfolderGranularity) { newValue in
-                model.handleAutoUploadSubfolderGranularityChange(newValue: newValue)
+            .tint(Color(NCBrandColor.shared.getElement(account: model.session.account)))
+            .onChange(of: model.autoUploadFull) { newValue in
+                albumModel.populateSelectedAlbums()
+                model.handleAutoUploadChange(newValue: newValue, assetCollections: albumModel.selectedAlbums)
             }
+            .font(.headline)
+            .toggleStyle(.button)
+            .buttonStyle(.bordered)
```
i2h3, line 184: Isn't there a common reusable SwiftUI view for this kind of button or is this here a unique appearance?

iOSClient/Settings/SelectAlbum/SelectAlbumView.swift
```diff
@@ -0,0 +1,146 @@
...
+    @Binding var selectedAlbums: Set<String>
+
+    var body: some View {
+        Section(NSLocalizedString("_albums_", comment: "")) {
+            ForEach(model.userAlbums, id: \.localIdentifier) { album in
+                SelectionButton(album: album, isSmartAlbum: false, selection: $selectedAlbums)
+            }
+        }
+    }
+}
```
i2h3, line 78: These two types are redundant and can be replaced by a single parameterized one, I guess.

iOSClient/Settings/SelectAlbum/SelectAlbumView.swift
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+        .foregroundColor(.primary)
+        .onAppear {
+            loader.loadImage(from: album, targetSize: .zero)
+        }
+    }
+}
+
+@MainActor class ImageLoader: ObservableObject {
```
i2h3, line 124: I would outsource this rather generic facility into a dedicated source code file to increase the visibility outside of this specific use case.

i2h3 (changes requested):
I have not tested it functionally yet.

mpivchev:
@i2h3 Did you manage to check functionality?

iOSClient/Networking/NCAutoUpload.swift
mpivchev: You are correct, will update

iOSClient/Settings/AutoUpload/NCAutoUploadView.swift
mpivchev: Models are instantiated from outside, so this is correct access control

iOSClient/Settings/AutoUpload/NCAutoUploadView.swift
i2h3: No, I meant within our own code in the project. Don't we have similar buttons elsewhere?

i2h3:
# Layout Improvements Needed

Screenshots taken on iPad 11" in landscape orientation.

## Automatic Uploads

* The scrollable content does not have any padding to the navigation bar.

Automatic Uploads

## Album Selection

* The scrollable content does not have any padding to the navigation bar.
* The navigation title is not utilized.

Album Selection

i2h3:
# Settings Not Reenabled After Uploads Completed

I waited for the automatic uploads of my favorites album to complete. The settings controls are not reenabled after completion. I wanted to also enable automatic upload of videos which I cannot. Navigating back and forth in the settings or selecting different tabs in the app does not help.

I first need to tap the stop button for the automatic uploading. That was not clear to me.

The German footer of the button "Obenstehende Optionen vor dem Hochladen anpassen." does not make much sense to me. The English reference is "Adjust the options above before uploading.". Either the translators need more context or the string needs to be more expressive.

I am a bit irritated by the necessity of manually starting and stopping the automatic upload which then I would not consider fully automated anymore. I would have assumed the upload actually is already enabled by the switch on the top.

i2h3:
# Missing Top Padding in Portrait Orientation on iPhone

Portrait
Landscape

i2h3:
# Target Folder Selection

In the rather verbose German language the selection for the destination folder becomes out of hand in terms of layout. I would have made it much simpler. Also, the label is too ambiguous. Even in English. "Select the auto upload folder" - what does that mean, exactly? What folder? What for? On device or on server? Since the overall context of automatic upload is clear, a plain "destination" appears more clear to me.

Suggestion

```swift
Section {
    LabeledContent {
        Text("/Photos")
    } label: {
        Text("Destination")
    }
}
```

Destination

i2h3:
# Unresponsive Application On Upload Start

I selected to upload photos and videos automatically via Wi-Fi only. Also I enabled the switch to only upload favorites. I did not select specific albums but left the default of the whole photo library. I selected to create subfolders by day. There are close to 12000 items in my photo library and I have iOS set to retrieve content on demand instead of having everything available all the time on device in original quality. When I start, the circular progress indicator appears and the app user interface stops being response. By accident I noticed that after some time the message of creating the photo archive appears. I just left the app open because I was writing the previous comment.

Freeze1
Freeze2

i2h3:
The upload itself, when it completed, uploaded what I expected and worked fine otherwise.

mpivchev:
> # Settings Not Reenabled After Uploads Completed
> I waited for the automatic uploads of my favorites album to complete. The settings controls are not reenabled after completion. I wanted to also enable automatic upload of videos which I cannot. Navigating back and forth in the settings or selecting different tabs in the app does not help.
> 
> I first need to tap the stop button for the automatic uploading. That was not clear to me.
> 
> The German footer of the button "Obenstehende Optionen vor dem Hochladen anpassen." does not make much sense to me. The English reference is "Adjust the options above before uploading.". Either the translators need more context or the string needs to be more expressive.
> 
> I am a bit irritated by the necessity of manually starting and stopping the automatic upload which then I would not consider fully automated anymore. I would have assumed the upload actually is already enabled by the switch on the top.

This work as intended. Before, the settings would be enabled, but changing them would not do anything until you stopped and started the auto uploaded again. Now is the same, as the upload functionality has not been changed (out of scope), but the settings are disabled so users do not get confused and wondering why nothing updates. 

I think generally it makes sense to disable the settings while auto upload is started, and I do not agree that it is not fully automated: it is automated for the settings chosen beforehand.

If you check Dropbox, it does pretty much the same things.

mpivchev:
> # Target Folder Selection
> In the rather verbose German language the selection for the destination folder becomes out of hand in terms of layout. I would have made it much simpler. Also, the label is too ambiguous. Even in English. "Select the auto upload folder" - what does that mean, exactly? What folder? What for? On device or on server? Since the overall context of automatic upload is clear, a plain "destination" appears more clear to me.
> 
> Suggestion
> 
> ```swift
> Section {
>     LabeledContent {
>         Text("/Photos")
>     } label: {
>         Text("Destination")
>     }
> }
> ```
> 
> Destination

Out of scope of this, but I agree with this. I will see how I can update it.

iOSClient/Networking/NCAutoUpload.swift
```diff
@@ -210,19 +214,49 @@ class NCAutoUpload: NSObject {
...
+        let assets: PHFetchResult<PHAsset> = PHAsset.fetchAssets(in: assetCollection, options: fetchOptions)
+        var assetResult: [PHAsset] = []
+
+        if selector == NCGlobal.shared.selectorUploadAutoUpload,
+           let idAssets = self.database.getPhotoLibraryIdAsset(image: tableAccount.autoUploadImage, video: tableAccount.autoUploadVideo, account: account) {
+            assets.enumerateObjects { asset, _, _ in
+                var creationDateString = ""
+                if let creationDate = asset.creationDate {
+                    creationDateString = String(describing: creationDate)
+                }
```
i2h3, line 227: Isn't this a use case for which optionals are there for?
```suggestion
                var creationDateString: String?

                if let creationDate = asset.creationDate {
                    creationDateString = String(describing: creationDate)
                }
```

iOSClient/Settings/AutoUpload/Albums.swift
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: Nextcloud GmbH
+// SPDX-FileCopyrightText: 2024 Milen Pivchev
```
i2h3, line 2:
```suggestion
// SPDX-FileCopyrightText: 2025 Milen Pivchev
```

iOSClient/Settings/AutoUpload/NCAutoUploadView.swift
i2h3: You mean injected? Okay, then it is fine. But on the other hand I would question why the other properties are `private` then. That was what my inconsistency was about. It is all inside the same and our module.

iOSClient/Settings/SelectAlbum/ImageLoader.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  ImageLoader.swift
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 27.01.25.
+//  Copyright © 2025 Marino Faggiana. All rights reserved.
+//
```
i2h3, line 7: Wrong license header.

i2h3 (changes requested):
There were some whitespace violations introduced, too, which need to be fixed.

Currently, I cannot repeat functional testing due to time constraints and priorities.

PR #3270, opened 2025-01-13
revert 3184-download-limit

marinofaggiana:


i2h3 (approved):
I will revisit the original branch once I am done with the reimplementation to also salvage in example the written documentation comments which might get lost otherwise.

PR #3280, opened 2025-01-23
Localization changes (Minor)

mpivchev:


mpivchev:
@rakekniven would you mind checking this? Thanks :)

mpivchev:
> On what basis did you make the changes here?

Generally for longer sentences it's good to have periods at the end. It's more professional IMO.

> We currently have more than 50 languages that have a translation rate of more than 50% of the ios app, which means that placing a period in a string invalidates 50 strings.

But aren't minor adjustments generally acceptable? Personally I can also update the periods for the sentences in all other languages in the project iself.

 Generally I would prefer to keep the periods, but if it can cause big issues then we can omit them. What do you think?

I will also update some strings that need more major changes (bad grammar, broken english, unclear etc.) and for those what do you suggest? Do I create completely new keys for them or reuse the old keys?

mpivchev:
> I cannot answer this as mentioned before as I do not speak/write asian/african/arabian and many other languages.
> 
> >

By "break something" I meant something in Transifex itself. Say, it notifies people of changes that must be made to periods even if the periods are already added manually. Personally I've only ever changed/added the default `en` strings in the iOS app, so I'm not sure what happens if I change a string from the other languages.

As a middle-ground would it be ok to keep the periods for English only, as that is (I assume) the most used language, and not add them to the other languages?

> Still my vote.

I know periods may be trivial but IMO it gives a more presentable and professional look for the app if big sentences have proper punctuation. As an example, all iOS strings in 1st party Apple apps that I have seen have periods at the end.

Issue #3291, opened 2025-01-30
Migrate to self-hosted Runners


i2h3:
As discussed with @tobiasKaminsky, I set up this issue for the idea of migrating our CI to a self-hosted macOS runner. It will be extended over the coming weeks to be discussed with all iOS development team members.

## Advantages

- **Drastic performance improvements** - The hosted GitHub actions macOS runners are very slow and in virtual machines. Running builds and tests on bare metal devices is a vast improvement in run time. In example, the `3184-download-limit2` actions took more than 30 minutes to complete. Clean building and running UI tests on a M3 MacBook Air take a bit more than 9 minutes. This is not a surprise when taking a look at [the specifications of the GitHub runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). It is surprising that Xcode and Simulator even launch with that amount of RAM.
- **Docker** - We can deploy Docker and have images cached for close to instant and throw-away deployments of Nextcloud test servers. In example the [Nextcloud Docker image](https://github.com/nextcloud/docker). Instead of a bare metal installation on macOS, like currently happening, it would be its conventional environment and isolated from interference with the host.
- **Inspection** - We can actually _see_ what is going on on the Mac. We do not have to guess based on command-line output.

## Disadvantages

- **Attack vector** - Public repositories are subject to eased introduction of malicious code into corporate networks. See [GitHub documentation](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security).
- **Increased Maintenance** - Someone has to take care of the runner and its integration. Though, it is not much additional work. I have experience with maintaining macOS CI runners for GitLab. Hence it likely will be me, as also discussed with @tobiasKaminsky. I will ensure that I won't become a knowledge silo and document it properly for everyone.

i2h3:
@marinofaggiana @mpivchev @Ivansss and @SystemKeeper might want to comment.

mpivchev:
This is amazing, hopefully it can work out :)

mpivchev:
> Increased Maintenance - Someone has to take care of the runner and its integration. Though, it is not much additional work. I have experience with maintaining macOS CI runners for GitLab.

[...]

PR #3311, opened 2025-02-14
Closes #3184: Download Limit Support

### LINKED ISSUE #3184: Download Limit Support

i2h3:
See commit message for details.


README.md
```diff
@@ -74,29 +74,48 @@ Do you want to try the latest version in development of Nextcloud iOS ? Simple,
...
-### Integration tests:
-To run integration tests, you need a docker instance of a Nextcloud test server. [This](https://github.com/szaimen/nextcloud-easy-test) is a good start.
+### Integration Tests
 
-1. In `TestConstants.swift` you must specify your instance credentials. App Token is automatically generated.
+In `TestConstants.swift` you must specify your credentials for the server you want to test against.
+[The Nextcloud Docker image](https://github.com/nextcloud/docker) maintained by the community is an easy and fast way to spin up a Nextcloud server to test against. 
+[This shallow CI server image](https://github.com/nextcloud/docker-ci/pkgs/container/continuous-integration-shallow-server) is also used by the Android project for automated tests in the GitHub pipeline.
+This latter one will be referred to from here on.
+With the following commands, you have will set up a new Nextcloud available on local HTTP port 8080, once the Docker image was initially pulled.
```
mpivchev, line 89:
```suggestion
With the following commands, you would set up a new Nextcloud Docker instance available on local HTTP port 8080.
```

Tests/NextcloudUITests/XCUIElement+await.swift
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: Nextcloud GmbH
```
mpivchev, line 1: Can we add this to an Extensions folder?

Tests/NextcloudUITests/XCUIElement+awaitInexistence.swift
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: Nextcloud GmbH
```
mpivchev, line 1: Can we add this to an Extensions folder?

Also, we can combine all these into one XCUIElement+Extensions file IMO. Sometimes there is such a thing as TOO much seperation.

Tests/NextcloudUITests/XCUIElement+awaitOrFail.swift
```diff
@@ -0,0 +1,37 @@
+// SPDX-FileCopyrightText: Nextcloud GmbH
```
mpivchev, line 1: Same here

iOSClient/Menu/NCShare+Menu.swift
```diff
@@ -54,6 +55,15 @@ extension NCShare {
...
                     advancePermission.metadata = self.metadata
+
+                    if let downloadLimit = try? self.database.getDownloadLimit(byAccount: self.metadata.account, shareToken: share.token) {
+                        advancePermission.downloadLimit = .limited(limit: downloadLimit.limit, count: downloadLimit.count)
+                    }
+
+                    if let downloadLimit = try? self.database.getDownloadLimit(byAccount: self.metadata.account, shareToken: share.token) {
+                        advancePermission.downloadLimit = .limited(limit: downloadLimit.limit, count: downloadLimit.count)
+                    }
+
```
mpivchev, line 66: This was written twice?

mpivchev (commented):
Some points. Will test functionality now

mpivchev:
This PR is quite big: 4 in 1. I would suggest to split something like this into smaller PRs for easier reviews.

mpivchev:
```
var isUiTestingEnabled: Bool {
        return ProcessInfo.processInfo.arguments.contains("UI_TESTING")
    }


    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        if isUiTestingEnabled {
            NCAccount().deleteAllAccounts()
        }
 ```
 
 Currently if the app already has an account the UI tests fail. You can use this bool when u prepare tests to delete all accounts.

File: Tests/Common/TestConstants.swift [hunk 6]
```diff
@@ -1,31 +1,44 @@
...
+    /// The default number of seconds to wait for the appearance of user interface controls during user interface tests.
+    ///
+    static let controlExistenceTimeout: Double = 60
+
+    ///
+    /// Unknown purpose.
+    ///
+    /// > To Do: Figure out whether this is obsolete or not.
+    ///
     static let timeoutLong: Double = 400
```
mpivchev, line 23: It was just a long timeout to wait for the server to do things, i see we don't use it so u can remove it

README.md
```diff
@@ -74,29 +74,48 @@ Do you want to try the latest version in development of Nextcloud iOS ? Simple,
...
+docker run \
+    --detach \
+    --name $CONTAINER_NAME \
+    --publish 8080:80 \
+    ghcr.io/nextcloud/continuous-integration-shallow-server:latest
+
+# Enable File Download Limit App.
+docker exec $CONTAINER_NAME su www-data -c "git clone --depth 1 -b $SERVER_VERSION https://github.com/nextcloud/files_downloadlimit.git /var/www/html/apps/files_downloadlimit/"
+docker exec $CONTAINER_NAME su www-data -c "php /var/www/html/occ app:enable files_downloadlimit"
 ```
```
mpivchev, line 110: As discussed you can just create this .sh file and include it in Tests. And you can also update the README explaining how to run it etc.

.github/workflows/xcode.yml
```diff
@@ -67,15 +72,22 @@ jobs:
         # Temporary workaround for missing pcntl_* in PHP 8.3: ini-values: apc.enable_cli=on
         ini-values: apc.enable_cli=on, disable_functions=
```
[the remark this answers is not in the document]
i2h3: I guess not. It was just copied from the Talk for iOS project.

.github/workflows/xcode.yml
```diff
@@ -84,18 +96,33 @@ jobs:
...
+
+    - name: Check if Safari is installed
+      run: |
+        if xcrun simctl listapps booted | grep -q com.apple.mobilesafari; then
+          echo "Safari is installed in the simulator."
+        else
+          echo "Safari is NOT installed in the simulator."
+          exit 1
+        fi
+
```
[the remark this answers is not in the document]
i2h3: According to the Talk for iOS project, it can happen to not be present. On iPhone as in Simulator, Safari can be removed while MDM can also remove it. This check was also copied from Talk for iOS project. I assume it is there for a reason. 😅

.github/workflows/xcode.yml
```diff
@@ -84,18 +96,33 @@ jobs:
...
-    - name: Check server status
+    - name: Assert Nextcloud Server Status
       run: curl -s --retry 5 --retry-delay 60 --retry-all-errors http://localhost:8080/status.php || true
 
-    - name: Test Nextcloud iOS
+    - name: Boot iOS Simulator
+      run: |
+        xcrun simctl boot "iPhone 16"
+        xcrun simctl bootstatus "iPhone 16" || echo "Simulator booted"
+
```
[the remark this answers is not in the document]
i2h3: The tests no longer create their own simulators. The default behavior was restored to use the simulators which are available. The now disabled test parallelization probably caused it. This is just an additional step for debugging GitHub actions because we otherwise do not know what is going on or whether the Simulator is already ready. Also copied from Talk for iOS as discussed with Ivan.

iOSClient/Menu/NCShare+Menu.swift
i2h3: This might have been a rebase accident. Over the weeks there were several dozen commits to catch up with.

i2h3:
> 4 in 1

Not that much. 😅 In regards to feature and the tests belonging to it: splitting those is not even what I know as "ready for review". Tests are part of a feature, not a distinct one.

The review amendments mentioned cannot be separated either. Before my initially integrated implementation got completely reverted as discussed with Marino, he did a review. I was taking his remarks into account with this new implementation. There was no previous feature implementation to amend right now. Maybe I should have just left that out for clarity.

The sharing UI improvements are changes along the way of implementation I did to better understand what is going on or enable the feature.

Tests/NextcloudUITests/XCUIElement+awaitInexistence.swift
i2h3: Based on my experience, the single extension works in this case but will cause a mess as soon as types get extended with capabilities belonging to different domains. "Melting pot" extensions of that kind have a high risk of getting messed up with dead code and difficult to read.

On the other hand, a new folder for a single item is questionable like a bullet point list with a single item.

I will do so, regardless.

Tests/Common/TestConstants.swift
i2h3: It is referenced by the dead integration test code of which there is only one commented out case for 2 years. I will remove that as well. Due to NextcloudKit changes in regards to Swift 6, modern concurrency and the introduction of Swift Testing it might be appropriate to discard those few comment lines and just start from scratch when we get to it.

mpivchev (commented):
All UI tests pass and functionality tested

i2h3:
> You can use this bool when u prepare tests to delete all accounts.

Just this was missing in the UI test implementation where I am launching the app:

```swift
app.launchArguments = ["UI_TESTING"]
```

mpivchev:
> > You can use this bool when u prepare tests to delete all accounts.
> 
> Just this was missing in the UI test implementation where I am launching the app:
> 
> ```swift
> app.launchArguments = ["UI_TESTING"]
> ```
> 
> But there is a problem with this: the app is launched for every test case. This means that the accounts are removed for every test case and the sign-in also happens again for every test case. This slows down UI tests a lot.

hmm, maybe you can use
```
#if Xcode && MANIFEST_LOADED_BY_XCODE
#endif
```

to only run this if running the tests through xcode. Not sure if it work, but worth a try?

mpivchev:
Or maybe you can have a flag in your test file that only allows to run the account deletion once?

i2h3:
> Or maybe you can have a flag in your test file that only allows to run the account deletion once?

I think it is fine. Actually, it is besser. It increases test isolation. Now the cherry on top would be to always deploy a fresh server container for every test.

i2h3:
I do not understand why but GitHub actions take horribly much time again (the test job about an hour), every test fails and everything just works fine locally. 😞

i2h3:
I have disabled the GitHub actions test job in the latest amendment.

* There are no other tests in place than the UI tests of this feature branch.
* UI tests are not feasible on GitHub hosted action runners due to their insufficient reliability and performance which stems from meager (virtualized) hardware specifications.
* Reduced feedback delay for pushed changes.
* Can be turned back on once we have unit tests or a proper UI test solution again (see updated #3291).

i2h3:
Oh… I should have squashed before merging. 🤦🏻‍♀️

PR #3325, opened 2025-02-18
Use correct User-Agent for login API requests

s-kaliberda:


mpivchev (approved):
Works great, thanks :)

Issue #3241, opened 2024-12-16
[Assistant_iOS] Improve Assistant on mobile screen


Hyeyoung346:
Please also check the current Android suggestions. It would be nice if iOS and Android have similar advancements and features. 
If you have other suggestions, tell me please. @mpivchev @tobiasKaminsky 

_**List view**_ 

- [x] Title box in the list view, perhaps it might be better to standardize.

=> Title: limit within 1 sentence, content: limit within 1 sentence
=> Content text & date text: I think it would be better to use both `Labels/Secondary` color.  
      Primary color looks a bit too strong to use for all texts. 
=> Date text: Make it a bit smaller, use `Callout/Regular`

- [x] Replace “Completed” to the showing the date as it is confusing which question was the oldest/latest

 => Check current Android screen that Alper shared. ex. 2 hours ago, 23 hours ago, Dec 12....
 => Keep “Pending” for pending tasks instead of timestamp 

<img width="452" alt="Image" src="https://github.com/user-attachments/assets/d310ed54-02e7-4229-803c-ff88996aaf33" />

******************************
_**Input / Output screen**_

- [x] I suggested switching to the conversational interface so that all “input” “output” labels can also be removed later. Keep current input/output screen for now, and then later can be further improved.

- [x] Add “Share” “Copy” “Edit” functions like Android 

I was thinking it would be nice if we could make these options available in the list view and input/output screen when users tap the input/output field. 
On the input/output page, I’d like it to look like this. But, if it’s too difficult to implement because of limited Assistant capabilities, we can add these functions later. 
Input: User tabs input box -> copy, select text
<img width="420" alt="Image" src="https://github.com/user-attachments/assets/3c9b94f1-2936-41dc-8e85-f71a6efc6929" />

Output: User tabs output box -> copy, select text, retry, share 
<img width="405" alt="Image" src="https://github.com/user-attachments/assets/a3b236de-9f73-4da6-a615-d924b00b0d47" />

 

- [x] Remove “Completed” state and displaying the date is also not essential in the input / output screen. We could add it later if many users require to add in the input/output page.
<img width="417" alt="Image" src="https://github.com/user-attachments/assets/554347fa-bd1a-42db-92fe-4d356e5f81db" />

Issue #3226, opened 2024-12-06
Assistant improvements


tobiasKaminsky:
List view
- Replace “completed” -> Display relative timestamp 12 hours ago, 1 week ago
- Keep "Pending" for pending tasks instead of timestamp  

Actions
- Copy / Share inside OS
- Edit question/request

PR #3327, opened 2025-02-19
Assistant design improvements + V2 API

### LINKED ISSUE #3201: [Assistant] A list view with scrolling issue
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### 
**### Actual behaviour**
When I scroll up, the list view is displayed all the way to the end of the bottom sheet. 

Image

### 
**### Expected behaviour**
The list view should not go any further until option button areas.

Image

mpivchev:
Prerequisite: https://github.com/nextcloud/NextcloudKit/pull/124

1. Implemented #3241 + #3226. 
Some points @Hyeyoung346:
- Tasks created in the same day show up as "Today" in the timeframe.  
- I think it's good to keep the status in the task detail, as just icons is not enough information IMO. In that case in the detail the status is shown like this:

image

- "Pending" does not show in the list and only shows on the detail, as that is where we get more information.

2. Fixed #3201 
3. Made Assistant compatible with API V1 (<NC30) and API V2 (>=NC30)
4. Added UI tests
      -  @i2h3 @marinofaggiana  You can run the UI tests to test V2 functionality. You need a server instance with assistant and testing app enabled. `Server.sh` now includes the commands to enable that, so you can start a server using that script.
      - For testing V1, please change the useV2 flag to false and test manually.


Hyeyoung346:
@mpivchev I think AI conversations are rarely “completed”. People often want to ask follow-up questions. This aligns better with the natural flow of human reasoning (where follow-up questions are common), therefore, should avoid “completed”mark. 

* Tasks created in the same day show up as "Today" in the timeframe.
=> I think "Today" is possible, but how to display the time frame after today in the list view?


* I think it's good to keep the status in the task detail, as just icons is not enough information IMO.
* "Pending" does not show in the list and only shows on the detail, as that is where we get more information.
=>  I think it would be useful to show the status (timeframe, pending) in the list view rather than in the task view. 
Maybe if it’s possible to implement follow-up questions with input filed inside the task detail later, probably would be useful to add good/bad feedback or report legal issue, modify response etc.


What is the difference between 	API V1 (<NC30) and API V2 (>=NC30)? Is there any improvements or difference from API V1?

mpivchev:
1. We can put the completion status on the left side, and the date on the right side?
I think in general having the icon on the list items is good enough for information, and it's not too cluttered. If you do want more information you can then open the details. 

I do think in general if we are going to have a message-like input and output structure we will have to redesign a lot of this. For example instead of straight up showing the status we can do it like a message saying "I could not generate a response, try again" or something similar. 

For now, however, since we just have input-output prompts, this is informative.

2. You didn't comment on this, so not sure if you agree.

3. As mentioned above, i think it becomes too cluttered.

At the moment it looks like this:

Hyeyoung346:
@mpivchev 
1. You mean `icon + date` in the list view? I think current list view screen looks good & informative. Perhaps the date could be displayed shorter => 25.Feb 2025.  
I mean `icon + date + completed` mark in the task details screen looks still questionable. I noticed that output responses sometimes also ask the following questions like “Would you like to know more about ~?”, so it looked strange to show the “completed” mark. Not sure `icon + date + completed` mark would be informative in the task detail screen, so I want to suggest to remove this part. But, as you said, we just have simple input-output prompts, so can iterate later further. 

2. You mean => Fixed https://github.com/nextcloud/ios/issues/3201
 Yeah if you fixed that scrolling issue in the list view, that would be awesome! 

Thanks for the detailed info! It looks better and more useful now, with more functionalities.

mpivchev:
Hi @Hyeyoung346 the scrolling issue is already fixed in this PR. It's just open until this PR is merged :)

mpivchev:
Regarding 1. I'll see what I can do on Monday. Have a nice weekend 😀

mpivchev:
@Hyeyoung346 modified the date to show in a more human readable way 
     - Less than a minute: Returns "Less than a minute ago".
     - Less than an hour: Returns the number of minutes (e.g., "5 minutes ago").
     - Less than a day: Returns the number of hours (e.g., "2 hours ago").
     - Less than a month: Returns the number of days (e.g., "3 days ago").
     - More than a month: Returns the full formatted date (e.g., "Jan 10, 2025").
     
image

> I mean icon + date + completed mark in the task details screen looks still questionable. I noticed that output responses sometimes also ask the following questions like “Would you like to know more about ~?”, so it looked strange to show the “completed” mark. Not sure icon + date + completed mark would be informative in the task detail screen, so I want to suggest to remove this part. But, as you said, we just have simple input-output prompts, so can iterate later further

I agree ultimately we should remove the status altogether and mimic a chat-like behavior. But for now, since it's just a prompt-style, we should keep it this way IMO.

mpivchev:
@jancborchardt can you quickly go over this :)

jancborchardt (changes requested):
Looks like a really nice improvement @mpivchev! :)

I only have 2 small details for now, but they are not necessarily blockers, could also be done in follow-ups:

- The vertical space between the 3 lines in the task list should be even (right now there is more space above the date)
- I agree the "Completed" text could be cut. The icon is fine enough. And in the detail view, if it is still pending or if there is an error, that should show more presently as a sort of empty content view where the Output would normally be.

i2h3:
How do I test this? Either with our corporate production instance or the test server run by the shell script I have eternally scheduled tasks but never responses.

mpivchev:
> Reference

image

this doesn't happen for me

mpivchev:
> @mpivchev Seems do not use all width and the font seems a bit too big
> 
> Screenshot 2025-02-27 alle 17 29 30

This doesnt happen for me

image

mpivchev:
> * The vertical space between the 3 lines in the task list should be even (right now there is more space above the date)

I feel like the text is too close together now. I would prefer to leave the padding @jancborchardt 

image

mpivchev:
image

This looks better

jancborchardt (commented):
@mpivchev can’t see your comment about the vertical spacing anymore, but if you think it’s too close together, let’s at least go for a padding value inbetween the current state and your screenshot. :)

mpivchev:
> @mpivchev can’t see your comment about the vertical spacing anymore, but if you think it’s too close together, let’s at least go for a padding value inbetween the current state and your screenshot. :)

Hi, added equal padding so it looks better now :) 

image

jancborchardt (approved):
Looking good, thank you! :)

i2h3:
@mpivchev Is fixing the translation form out of scope? There is an obvious mismatch between the web version and what the app offers with its single text field and no hint what to enter there.

App
Web

Tests/NextcloudUITests/AssistantUITests.swift
```diff
@@ -0,0 +1,151 @@
...
+        goToAssistant()
+
+        createTask(input: taskInputCreated)
+
+        pullToRefresh()
+
+        try await aMoment()
+
+        let cell = app.collectionViews.children(matching: .cell).element(boundBy: 0)
+        XCTAssert(cell.staticTexts[taskInputCreated].exists)
```
i2h3, line 103: This fails for me when testing against tech preview because "an error occurred" when creating the task, so there is no cell to find.

Tests/NextcloudUITests/AssistantUITests.swift
```diff
@@ -0,0 +1,151 @@
...
+
+        let talkStaticText = app.tables.staticTexts["Assistant"]
+        talkStaticText.tap()
+    }
+
+    private func createTask(input: String) {
+        app.navigationBars["Assistant"].buttons["CreateButton"].tap()
+
+        app.textViews["InputTextEditor"].typeText(input)
+        app.navigationBars["New Free text to text prompt task"].buttons["Create"].tap()
```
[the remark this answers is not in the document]
i2h3: Interestingly, this is different when testing against the server on `localhost`:
```plain
([String]) 2 values {
  [0] = "More"
  [1] = "New Free text to text prompt task"
}
```

Tests/NextcloudUITests/AssistantUITests.swift
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    // MARK: - Tests
+
+    func testCreateAssistantTask() async throws {
+        goToAssistant()
+
+        createTask(input: taskInputCreated)
+
+        pullToRefresh()
```
i2h3, line 98: When running against localhost, a short delay is needed in between, otherwise pull to refresh will fail because the cell has not appeared yet in the visible area.

i2h3 (changes requested):
UI tests are fragile and fail for me. Mostly it is the login and not your fault. By now I think we should replace the fragile web view interaction by just injecting an app token through an app launch argument so the whole login can be skipped and tests run much faster to the point.

The tests did not work on the daily server, but tech preview somewhat better.

Also it was strange that I first had to remove all the unnecessary framework references from the Xcode UI test target. Xcode was complaining about RealmSwift not being found when linking. It is not even supposed to be in the UI test target. 😵‍💫

I stop at this point now, I could spend the rest of the day making the tests work.

mpivchev:
> UI tests are fragile and fail for me. Mostly it is the login and not your fault. By now I think we should replace the fragile web view interaction by just injecting an app token through an app launch argument so the whole login can be skipped and tests run much faster to the point.
> 
> The tests did not work on the daily server, but tech preview somewhat better.
> 
> Also it was strange that I first had to remove all the unnecessary framework references from the Xcode UI test target. Xcode was complaining about RealmSwift not being found when linking. It is not even supposed to be in the UI test target. 😵‍💫
> 
> I stop at this point now, I could spend the rest of the day making the tests work.

The tests were made to be ran with `Server.sh` just like the download limit tests. Using any other server can cause the tests to fail as they may have different tasks enabled and so on.

i2h3:
> Regarding injecting the app token, i think we did that before, let me check.

I think Talk team does it.

iOSClient/Assistant/Models/NCAssistantModel.swift
```diff
@@ -0,0 +1,259 @@
...
+                return
+            }
+
+            self.tasks = tasks
+            self.filterTasks(ofType: self.selectedType)
+        }
+    }
+}
+
+extension NCAssistantModel {
```
i2h3, line 216: If I see correctly, this is used in SwiftUI previews only. Hence I would wrap it in `#if DEBUG`.

i2h3 (changes requested):
The UI tests now ran through now. 🤷🏻‍♀️ Both on iPhone and iPad simulator. With the requested changes it is fine merge then.

mpivchev:
@marinofaggiana please make sure to also test V1, with christine or daily nc instance.

mpivchev:
@i2h3 all PR suggestions added

i2h3 (approved):

PR #3351, opened 2025-03-10
Fix login poll (again) + Additional bug fix

mpivchev:
1. Fixes login poll (again)
2. Fix an issue happening when removing the last account and going back from log in would close both presented login modals and cause the app to go back to home screen with no account.


i2h3:
I will wait for https://github.com/nextcloud/NextcloudKit/pull/128 to be integrated into `develop` first.

I know that the fix for the login poll works already.

I am confused about the explicit NextcloudKit 6.0.2 reference here. The polling fix is not even on `develop`, so actually the development branch reference would have been appropriate (once the fix there is integrated).

PR #3356, opened 2025-03-12
Merge Develop into Master

i2h3:
This is just a test request to analyze what is wrong with the repository history.


mpivchev:
I think squishing PRs breaks the history. It's a well known consequence of squishing.

i2h3:
> I think squishing PRs breaks the history. It's a well known consequence of squishing.

Squashing feature branches before integration into `develop` is completely fine and actually there are projects and teams preferring that (I used to work in such and am on that side). There rarely are reasonable edge cases to stray from that process. It results in a much more clear version history to read later. It does not "break" history.

What actually breaks history is squashing all the commits from `develop` when integrating them into `master`. You loose history information. You cannot understand which commits and changes exactly are part of `master` then. Well, at least I cannot recognize it anymore. I see some familiar changes from `develop` but they all are in a "6.4.0" commit.

PR #3362, opened 2025-03-13
New assistant button

marinofaggiana:


mpivchev (changes requested):
1. 
image

The icon in Media is a bit too big and does not look so nice. I think making it a bit smaller would be good.

2. Icon should also show in favorites. It does not.

PR #3363, opened 2025-03-13
Remove old code for Nextcloud 17 or lower

mpivchev:
Closes https://github.com/nextcloud/files-clients/issues/79#issuecomment-2540908515


i2h3 (approved):
Where is the check for the server to be the oldest supported version?

PR #3386, opened 2025-03-27
Add auto login for Debug

mpivchev:
Adds an auto logging button for debugging so you can skip logging in manually every time.

Use XCode env vars to log in to any server.

image

image


mpivchev:
Need to use .plist instead or some other solution to be able to gitignore env vars

PR #3389, opened 2025-03-30
Update NCNetworking+WebDAV.swift

DaCHack:
Extending timeout for webdav folder reads to 5min to allow for Scanning folders with large number of files.

Editing and further testing by maintainers appreciated.

### DIFF
--- a/iOSClient/Networking/NCNetworking+WebDAV.swift
+++ b/iOSClient/Networking/NCNetworking+WebDAV.swift
@@ -56,7 +56,7 @@ extension NCNetworking {
                                              depth: "1",
                                              showHiddenFiles: NCKeychain().showHiddenFiles,
                                              account: account,
-                                             options: NKRequestOptions(queue: queue)) { task in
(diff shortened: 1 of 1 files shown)


DaCHack:
> @DaCHack Thanks for your PR, 5 minutes is really too much, however the pagination of the results is planned which will perhaps also solve this problem.

My 10,000-file photo folder syncing from iPhone takes ~3min to load on the webpage currently. This is why I used this value to at least have some headroom based on my rather average use case.

Can you share the issue or PR that tracks the pagination for the iOS app? Fine to leave it to this solution if this solves the problem.

In the meantime: What are the downsides of the 5min-workaround? Why are we timing out on iOS while there is no timeout on the web app? We could release the quick fix and revert back with the pagination PR...

DaCHack:
@mpivchev and @marinofaggiana 
Could You please Point me to the corresponding pagination PR or issue to see how I can help?

Or alternatively
A) accept this Change as a workaround
B) give architectural guidance on other ways to avoid the timeout

This issue is basically Rendering the App unusable

PR #3463, opened 2025-05-06
Realm-Queue-refactoring

marinofaggiana:
1. performRealmRead<T>(_ block: (Realm) throws -> T?) -> T?

Purpose:
This function is designed to safely perform read operations from Realm within a serial queue. It uses the provided block of code to interact with the database and handles any read errors.

How it works:
	•	realmQueue.sync: Ensures that Realm operations are executed synchronously, preventing concurrency issues between reads.
	•	Block block(realm): This block is passed as an argument and is executed using the provided Realm instance. The block can return a value of generic type T?.
	•	Error handling: If an error occurs during the read (e.g., a database access issue), an error log is recorded, and the function returns nil.

Typical use case:
	•	Safe read operations from Realm: It’s used to retrieve data from Realm without worrying about managing queues or database access errors.

⸻

2. performRealmWrite(sync: Bool = true, _ block: @escaping (Realm) throws -> Void)

Purpose:
This function is used to perform write operations on Realm safely, allowing synchronization or asynchronous execution depending on the sync parameter.

How it works:
	•	Block executionBlock: The block that contains the write logic. This block is passed as a parameter and performs the write operation inside a Realm write transaction.
	•	realmQueue.sync or realmQueue.async: Depending on the value of sync, the operation is executed either synchronously (sync) or asynchronously (async). The write operation doesn’t block the UI thread, thanks to the execution queue.
	•	autoreleasepool: Ensures that temporary objects created during the write are properly deallocated.
	•	Error handling: If an error occurs during the write, an error log is recorded, but no exception is thrown. The app continues running without crashing.

Typical use case:
	•	Safe write operations to Realm: It’s used to perform write operations in a safe context, ensuring that concurrency is properly handled.
	•	Asynchronous or synchronous execution: The function allows you to choose whether to execute the write operation synchronously (where the calling thread waits for completion) or asynchronously (where the operation runs in the background).

⸻

Why these functions are useful:
	•	Safe synchronization of read and write operations: They prevent conflicts between simultaneous operations on Realm, ensuring that data is not read or written concurrently from different threads.
	•	Automatic error handling: Errors are logged without causing app crashes, allowing them to be handled with logs and debugging, rather than stopping the app.
	•	Support for both asynchronous and synchronous operations: The function gives you the flexibility to choose how to perform operations based on the app’s needs, improving performance or UI responsiveness.

example: 

```
func addMetadatas(_ metadatas: [tableMetadata]) {
        performRealmWrite { realm in
            for metadata in metadatas {
                realm.create(tableMetadata.self, value: metadata, update: .all)
            }
        }
 }
```

```
func getResultMetadata(predicate: NSPredicate) -> tableMetadata? {
        return performRealmRead { realm in
            realm.objects(tableMetadata.self)
                .filter(predicate)
                .first
        }
 }
```

### DIFF
--- a/iOSClient/Networking/NCNetworking+Upload.swift
+++ b/iOSClient/Networking/NCNetworking+Upload.swift
@@ -297,6 +297,7 @@ extension NCNetworking {
                         size: Int64,
                         task: URLSessionTask,
                         error: NKError) {
+        isAppSuspending = false
 
 #if EXTENSION_FILE_PROVIDER_EXTENSION
 
@@ -361,7 +362,6 @@ extension NCNetworking {
                         date: Date?,
                         size: Int64,
                         error: NKError) {
-        isAppSuspending = false
 
         DispatchQueue.main.async {
             var isApplicationStateActive = false


mpivchev:
Thanks for the extensive explanation <3 @marinofaggiana

i2h3 (approved):

PR #3477, opened 2025-05-15
Auto upload - refactoring

marinofaggiana:


philiprenich:
@marinofaggiana Does this solve the issue of duplicate uploads for users who turn back on auto-upload (but without the "only new from today" option) if they had their auto-upload disabled by the previous app update?

That doesn't sound clear, so written another way, will the auto-uploader now check for photos/videos that already exist in/have been uploaded to the Destination folder?

PR #3500, opened 2025-06-03
Handle permissions

mpivchev:
Implements https://github.com/nextcloud/files-clients/issues/22


mpivchev:
Currently on iOS files in file provider cannot be directly modified in 3rd party apps like Word or Pages, as we use the old `NSFileProviderExtension` instead of the new `NSFileProviderReplicatedExtension` available for iOS16+. The old implementation uses virtual files instead of a full sandbox representation of files, which makes 2 way sync not possible for 3rd party apps, as they rely on the actual files instead of the virtual ones.

Implementing `NSFileProviderReplicatedExtension` has been discussed between me and @claucambra and can be a potential feature.

@tobiasKaminsky @marinofaggiana 

More info here: https://developer.apple.com/documentation/fileprovider

mpivchev:
Moreover, with the new way of representing files (as a sandbox) we may lose the ability to show shared files in the file provider, due to privacy concerns or sync issues, but this must be investigated further

iOSClient/Main/Collection Common/NCCollectionViewCommon+CollectionViewDataSource.swift
```diff
@@ -141,7 +141,7 @@ extension NCCollectionViewCommon: UICollectionViewDataSource {
 
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
         var cell: NCCellProtocol & UICollectionViewCell
-        let permissions = NCPermissions()
+        let metadataPermissions = NCMetadataPermissions()
```
alperozturk96, line 144: Use static version of it.

iOSClient/Share/Advanced/NCShareCells.swift
```diff
@@ -123,10 +123,10 @@ enum NCLinkEmailPermission: CaseIterable, NCPermission {
 
     var permissionBitFlag: Int {
         switch self {
-        case .read: return NCPermissions().permissionReadShare
-        case .edit: return NCPermissions().permissionEditShare
-        case .create: return NCPermissions().permissionCreateShare
-        case .delete: return NCPermissions().permissionDeleteShare
+        case .read: return NCSharePermissions().permissionReadShare
```
alperozturk96, line 126: Maybe you can omit the return keyword from each case.

iOSClient/Share/NCMetadataPermissions.swift
```diff
@@ -0,0 +1,48 @@
...
+//  Nextcloud
+//
+//  Created by Milen Pivchev on 21.05.25.
+//  Copyright © 2025 Marino Faggiana. All rights reserved.
+//
+
+import Foundation
+
+/// Metadata permissions, represented as symbols (letters)
+class NCMetadataPermissions: NSObject {
```
alperozturk96, line 12: Make this struct and singleton if possible.

alperozturk96 (changes requested):

iOSClient/Share/Advanced/NCShareCells.swift
```diff
@@ -65,11 +65,11 @@ enum NCUserPermission: CaseIterable, NCPermission {
 
     var permissionBitFlag: Int {
         switch self {
```
alperozturk96, line 67: return switch here?

alperozturk96 (approved):

PR #3502, opened 2025-06-04
Added verbose logging to the login polling.

i2h3:
- Removed user-defined device name from user agent string because since iOS 16 only a generic one is returned for privacy reasons anway. This also makes the user agent string consistent with other requests from the iOS app so server logs can be filtered correctly.
- Updated web view to use a private browser session which does not persist any data on disk but only keeps them in memory for the lifetime of the web view.
- Updated variable names to remove redundancy and improve overall code legibility.
- Added a lot of logging statement to NCLoginProvider.
- Added some documentation comments on NCLoginProvider.
- Updated NextcloudKit reference to 6.0.10.

### DIFF
--- a/Brand/Intro/NCIntroViewController.swift
+++ b/Brand/Intro/NCIntroViewController.swift
@@ -189,7 +189,7 @@ class NCIntroViewController: UIViewController, UICollectionViewDataSource, UICol
     @IBAction func signupWithProvider(_ sender: Any) {
         if let viewController = UIStoryboard(name: "NCLogin", bundle: nil).instantiateViewController(withIdentifier: "NCLoginProvider") as? NCLoginProvider {
             viewController.controller = self.controller
-            viewController.urlBase = NCBrandOptions.shared.linkloginPreferredProviders
+            viewController.initialURLString = NCBrandOptions.shared.linkloginPreferredProviders
             self.navigationController?.pushViewController(viewController, animated: true)
         }
     }
--- a/iOSClient/Login/NCLogin.swift
+++ b/iOSClient/Login/NCLogin.swift
@@ -328,8 +328,9 @@ class NCLogin: UIViewController, UITextFieldDelegate, NCLoginQRCodeDelegate {
                 NextcloudKit.shared.getLoginFlowV2(serverUrl: url, options: loginOptions) { [self] token, endpoint, login, _, error in
                     // Login Flow V2
                     if error == .success, let token, let endpoint, let login {
+                        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Successfully received login flow information.")
                         let safariVC = NCLoginProvider()
-                        safariVC.urlBase = login
+                        safariVC.initialURLString = login
                         safariVC.uiColor = textColor
                         safariVC.delegate = self
                         safariVC.startPolling(loginFlowV2Token: token, loginFlowV2Endpoint: endpoint, loginFlowV2Login: login)
@@ -419,11 +420,15 @@ class NCLogin: UIViewController, UITextFieldDelegate, NCLoginQRCodeDelegate {
     }
 }
 
+// MARK: - NCShareAccountsDelegate
+
 extension NCLogin: NCShareAccountsDelegate {
     func selected(url: String, user: String) {
         isUrlValid(url: url, user: user)
     }
 }
+
+// MARK: - UIDocumentPickerDelegate
 
 extension NCLogin: ClientCertificateDelegate, UIDocumentPickerDelegate {
     func didAskForClientCertificate() {
@@ -465,6 +470,8 @@ extension NCLogin: ClientCertificateDelegate, UIDocumentPickerDelegate {
         }
     }
 }
+
+// MARK: - NCLoginProviderDelegate
 
 extension NCLogin: NCLoginProviderDelegate {
     func onBack() {
--- a/iOSClient/Login/NCLoginProvider.swift
+++ b/iOSClient/Login/NCLoginProvider.swift
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: Nextcloud GmbH
+// SPDX-FileCopyrightText: 2025 Iva Horn
 // SPDX-FileCopyrightText: 2025 Milen Pivchev
 // SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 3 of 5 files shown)


Brand/Intro/NCIntroViewController.swift
```diff
@@ -189,7 +189,7 @@ class NCIntroViewController: UIViewController, UICollectionViewDataSource, UICol
     @IBAction func signupWithProvider(_ sender: Any) {
         if let viewController = UIStoryboard(name: "NCLogin", bundle: nil).instantiateViewController(withIdentifier: "NCLoginProvider") as? NCLoginProvider {
             viewController.controller = self.controller
-            viewController.urlBase = NCBrandOptions.shared.linkloginPreferredProviders
+            viewController.initialURLString = NCBrandOptions.shared.linkloginPreferredProviders
```
mpivchev, line 192: We use `urlBase` all troughout the app. Not sure if only changing it here is a good idea.

iOSClient/Login/NCLogin.swift
```diff
@@ -419,12 +420,16 @@ class NCLogin: UIViewController, UITextFieldDelegate, NCLoginQRCodeDelegate {
...
 
+// MARK: - NCShareAccountsDelegate
+
 extension NCLogin: NCShareAccountsDelegate {
     func selected(url: String, user: String) {
         isUrlValid(url: url, user: user)
     }
 }
 
+// MARK: - UIDocumentPickerDelegate
```
mpivchev, line 431: Are these needed? It's good for general categorization but here it's a bit obvious what is coming after.

iOSClient/Login/NCLoginProvider.swift
```diff
@@ -7,91 +8,112 @@ import UIKit
...
         super.viewDidDisappear(animated)
+        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Login provider view did disappear.")
+
         NCActivityIndicator.shared.stop()
 
+        guard pollingTask != nil else {
+            return
+        }
+
+        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Cancelling existing polling task because view did disappear...")
```
mpivchev, line 97:
```suggestion
        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Cancelling existing polling task because Login view did disappear...")
```

iOSClient/Login/NCLoginProvider.swift
```diff
@@ -7,91 +8,112 @@ import UIKit
...
-        }
-
         request.addValue("true", forHTTPHeaderField: "OCS-APIRequest")
         request.addValue(language, forHTTPHeaderField: "Accept-Language")
 
         webView.load(request)
     }
 
+    ///
+    /// Dismiss the login web view from the hierarchy.
```
mpivchev, line 115: A bit obvious I think.

iOSClient/Login/NCLoginProvider.swift
```diff
@@ -145,32 +210,48 @@ class NCLoginProvider: UIViewController {
...
     }
 }
 
+// MARK: - WKNavigationDelegate
+
 extension NCLoginProvider: WKNavigationDelegate {
     func webView(_ webView: WKWebView, didReceiveServerRedirectForProvisionalNavigation navigation: WKNavigation!) {
-        guard let url = webView.url else { return }
-        let urlString: String = url.absoluteString.lowercased()
+        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Web view did receive server redirect for provisional navigation.")
```
mpivchev, line 222: You can store `NextcloudKit.shared.nkCommonInstance` in a variable so we don't write the whole thing every time

iOSClient/Login/NCLoginProvider.swift
```diff
@@ -145,32 +210,48 @@ class NCLoginProvider: UIViewController {
...
+
             self.present(alertController, animated: true)
+
             return
         }
 
-        // Login via provider
-        if urlString.hasPrefix(NCBrandOptions.shared.webLoginAutenticationProtocol) == true && urlString.contains("login") == true {
+        // Login via provider.
+        if currentWebViewURLString.hasPrefix(NCBrandOptions.shared.webLoginAutenticationProtocol) == true && currentWebViewURLString.contains("login") == true {
```
mpivchev, line 247: Do we need the `== true`?

iOSClient/Login/NCLoginProvider.swift
```diff
@@ -7,91 +8,112 @@ import UIKit
 import NextcloudKit
 import FloatingPanel
 
+protocol NCLoginProviderDelegate: AnyObject {
+    ///
+    /// Called when the back button is tapped in the login provider view.
```
mpivchev, line 13: A bit obvious.

iOSClient/Login/NCLoginProvider.swift
```diff
@@ -7,91 +8,112 @@ import UIKit
...
+        self.webView = webView
+
         let navigationItemBack = UIBarButtonItem(image: UIImage(systemName: "arrow.left"), style: .done, target: self, action: #selector(goBack(_:)))
         navigationItemBack.tintColor = uiColor
         navigationItem.leftBarButtonItem = navigationItemBack
     }
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Login provider view did appear.")
```
mpivchev, line 67: Personally not a big fan of the "did X" logs. "Login provider appeared" might be more suitable.

iOSClient/NCAccount.swift
```diff
@@ -58,19 +60,24 @@ class NCAccount: NSObject {
         NextcloudKit.shared.getUserProfile(account: account) { account, userProfile, _, error in
             if error == .success, let userProfile {
                 /// Login log debug
-                NextcloudKit.shared.nkCommonInstance.writeLog("[DEBUG] Create new account \(account) with user \(user) and userId \(userProfile.userId)")
+                NextcloudKit.shared.nkCommonInstance.writeLog(info: "Got user profile, creating new account \(account) with user \(user) and userId \(userProfile.userId)")
```
mpivchev, line 63: Here we don't use "did get user profile", but a more "human" response. I prefer this, and it's mostly how we log throughout the app.

iOSClient/NCAccount.swift
```diff
@@ -36,6 +36,8 @@ class NCAccount: NSObject {
                        password: String,
                        controller: NCMainTabBarController?,
                        completion: @escaping () -> Void = {}) {
+        NextcloudKit.shared.nkCommonInstance.writeLog(info: "NCAccount will create account...")
```
mpivchev, line 39:
```suggestion
        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Creating account")
```

iOSClient/NCAccount.swift
```diff
@@ -58,19 +60,24 @@ class NCAccount: NSObject {
...
                     NCKeychain().setClientCertificate(account: account, p12Data: NCNetworking.shared.p12Data, p12Password: NCNetworking.shared.p12Password)
+
                     if let controller {
                         controller.account = account
+                        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Dismissing login provider view controller...")
                         viewController.dismiss(animated: true)
 
                         completion()
                     } else if let controller = UIStoryboard(name: "Main", bundle: nil).instantiateInitialViewController() as? NCMainTabBarController {
+                        NextcloudKit.shared.nkCommonInstance.writeLog(info: "Presenting initial view controller from main storyboard...")
```
mpivchev, line 80: Is this needed to be logged? Seems a bit trivial.

mpivchev (changes requested):

iOSClient/Login/NCLogin.swift
i2h3: Comments are always optional. 😉 😅  Apparently you are not using the Xcode navigation bar which I rely on very much like keyboard navigation, especially in longer source code files of which we have plenty of. `⌃ + 5` (Show Group Files) and `⌃ + 6` (Show Document Items) are essential but list extensions without further details. These comments do not only introduce a divider in the outline but also add context to the extension. See screenshot below.

They are not need but a valuable convenience to contributors.

Bildschirmfoto 2025-06-05 um 10 30 52

iOSClient/Login/NCLoginProvider.swift
i2h3: I do not understand. We do not have a symbol named like that and [login](https://www.oxfordlearnersdictionaries.com/definition/english/login) is a standard noun which is not subject to capitalization in English.

iOSClient/Login/NCLoginProvider.swift
i2h3: I will change this because I already was annoyed by that.

iOSClient/Login/NCLoginProvider.swift
i2h3: Actually not at all (to me) because you first have to look into the implementation that is about the UIKit navigation controller hierarchy and not the WKWebView navigation. With this explicit documentation comment, contributors can always know directly by using Quick Help (inspector or popover through keyboard shortcut). I rely on this a lot, too.

iOSClient/NCAccount.swift
i2h3: Yes, this is what this whole pull request is about to gather as much information as possible during login to have a more complete picture of what is happening when. This is important for some customer support tickets.

iOSClient/Login/NCLoginProvider.swift
i2h3: As explained in the other response, in context of a `WKWebView` being part of the controller's managed view hierarchy it is ambiguous. I wrote it for a reason. 😁 Next time, just like any contributors, I do not have to look it up and read the implementation to be completely sure what it does.

iOSClient/Login/NCLoginProvider.swift
i2h3: I will update this according to your suggestion.

iOSClient/NCAccount.swift
i2h3: I will update this.

iOSClient/Login/NCLoginProvider.swift
i2h3: No, I did not look at that part actually, I just changed the variable name.

i2h3:
Changed target branch to `6.6.4-fix` as advised by @marinofaggiana to release independently from the unstable `master`.

iOSClient/Login/NCLogin.swift
mpivchev: I see. I tried using it before but could never stick with that 😄

iOSClient/Login/NCLoginProvider.swift
mpivchev: Fair enough 👍

mpivchev (approved):

PR #3560, opened 2025-06-30
brand-push-server

marinofaggiana:
Brand pushNotification Server required a value


Brand/NCBrand.swift
```diff
@@ -110,9 +110,14 @@ final class NCBrandOptions: @unchecked Sendable {
...
             }
         }
-#if DEBUG
+        #if DEBUG
         pushNotificationServerProxy = "https://c0004.customerpush.nextcloud.com"
-#endif
+        #else
+        if pushNotificationServerProxy.isEmpty,
+            brand == "Nextcloud" {
+            pushNotificationServerProxy = "https://push-notifications.nextcloud.com"
```
tobiasKaminsky, line 118: Hm. But this means for branded clients without their own push proxy, we still try to connect to our server.
And that is wrong. Since this will not work, we should not do it.

I think it is better to change the logic, that only a push connection is made, when there is a server (pushNotificationServerProxy.isNotEmpty)

tobiasKaminsky (changes requested):

PR #3564, opened 2025-07-01
Localized Download Limit User Interface

i2h3:


marinofaggiana:
closed, because the version now is >= 7

PR #3566, opened 2025-07-01
Fix folder naming not using capabilities renamer

mpivchev:
Naming a folder **A*** where the "*" is not part of the forbidden characters coming from capabilities, would cause the folder to be named **A** instead.

Fixed by making the folder creating take into account forbidden characters from capabilities instead of generic forbidden characters.

### DIFF
--- a/iOSClient/Extensions/UIAlertController+Extension.swift
+++ b/iOSClient/Extensions/UIAlertController+Extension.swift
@@ -101,7 +101,6 @@ extension UIAlertController {
             textField.autocapitalizationType = .words
         }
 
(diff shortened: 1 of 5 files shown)


i2h3 (approved):

i2h3:
Please add instructions how to test, though.

mpivchev:
> Please add instructions how to test, though.

Added

mpivchev:
@marinofaggiana did you test before merging?

marinofaggiana:
@mpivchev haven't you tried it? (it seems like a simple function, doesn't it?)

PR #3575, opened 2025-07-06
Add a CSV fallback for Nextcloud-Text

7408647n:
Add a fallback for CSV files if the editor is Nextcloud-Text, as the Nextcloud-Text editor displays CSV files as text rather than as a table.


mpivchev:
Hi @7408647n, can you please rebase with master? I can then test and merge your fix. Thanks :)

mpivchev:
Hi, I don't think we can merge this since we must use Nextcloud Text for files like this. I will check with the Nextcloud Text devs and see what we can do.

mpivchev:
In the meantime I can repurpose your code to open a csv in `QTViewer` only **IF** Nextcloud Text is not installed.

PR #3623, opened 2025-07-24
Versio 7.1.0 File provider multi domain

marinofaggiana:
- Available only from iOS 17


marinofaggiana:
@mpivchev New big version with multi domain FPE and for (minimun) iOS 17

PR #3685, opened 2025-08-18
feat: Rewrite mTLS logic

mpivchev:
- Use Keychain to save Identity instead of NCPreferences
- Delete old Identity in keychain before replacing
- Make NCLoginProvider also check for Identity

Update 1:
- Fixed private scope access of .p12 files on physical devices
- New screen for uploading certificate

image

TODO:
Test if Keychain can accept multiple different identities from a domain with different ports (ex localhost)
Test cert expiration


marinofaggiana:
good to remove the certificate to account and move for host !!

< pls rebase >

ReanimationXP:
does this commit mean mTLS certs can be imported and used for Cloudflare Tunnels / Zero Trust? do we know when this can be merged?

martymccasland:
I don't know if this helps, but the Immich iOS successfully implements mTLS certificates installed on iOS.   If that code is available, you could see what they've done that works.

As is, even pointing the NextCloud iOS app to the p12, it does not pass the certificate to the server.

mpivchev:
> I don't know if this helps, but the Immich iOS successfully implements mTLS certificates installed on iOS. If that code is available, you could see what they've done that works.
> 
> As is, even pointing the NextCloud iOS app to the p12, it does not pass the certificate to the server.

Hi @martymccasland. 
Immich also uses `UIDocumentPickerDelegate` to pick a certificate inside the app, as apps are sandboxed and cannot access iOS certificates. https://github.com/immich-app/immich/blob/b2050583f5dfcebf2c26808d446d0cc513c8c171/mobile/ios/Runner/Core/NetworkApiImpl.swift#L19C1-L26C4

I will investigate this further.

PR #3708, opened 2025-08-26
speed up sync of metadata

marinofaggiana:
when open app or user goes into a folder

- directly fetch newest metadata of this folder (first check via etag)
- then compare all subfolders for changed etag
- sync metadata of each non-updated subfolder one by one
- make sure that this operation is cancelled as soon as the user does something else (navigate away, clicks on a file, closes app, etc.)


iOSClient/Files/NCFiles+SpeedUpSyncMetadata.swift
```diff
@@ -0,0 +1,178 @@
+// SPDX-FileCopyrightText: Nextcloud GmbH
+// SPDX-FileCopyrightText: 2025 Marino Faggiana
```
mpivchev, line 2: Good to rename this file to MetadataSync.swift

PR #3719, opened 2025-08-29
Code improved

marinofaggiana:


Liaders:
The problem with cyrillic push notifications still persists in 7.1.3

marinofaggiana:
@Liaders yes, it has not been completed.

PR #3800, opened 2025-09-30
Make auto upload button fixed

mpivchev:
image

image

Also supports pre ios 26


mpivchev:
@kra-mo (Laura) approved it in terms of design

kra-mo (approved):

PR #3814, opened 2025-10-08
sync improved

marinofaggiana:
**Skips processing the current directory if it has not changed recently.**

The loop continues only when both the ETag matches the metadata (indicating no content change)
and the last synchronization date is within the last 3 hours.
This prevents redundant work on directories that were already synced recently.

- Conditions:
- `directory?.etag == metadata.etag`: the directory content has not changed.
- `let lastSyncDate = directory?.lastSyncDate as? Date`: ensures the last sync date exists and is convertible to `Date`.
- `lastSyncDate >= Date().addingTimeInterval(-3 * 3600)`: skips only if the last sync occurred less than 3 hours ago.

In all other cases (different ETag, missing date, or sync older than 3 hours), the loop proceeds normally.


marinofaggiana:
@tobiasKaminsky @alperozturk96 ( as discussed in Berlin )

PR #3840, opened 2025-10-27
LucidBanner

marinofaggiana:
**A new library UI for banner.**

Simulator Screenshot - iPhone 17 - 2025-10-27 at 15 53 40


marinofaggiana:
@mpivchev please make test !

mpivchev:
Some things that can be a problem:

- **Potential issues with the banner hiding things behind it?** For example in Files it hides some ... files.... when its showing, or its hiding the account switcher, or the title. i discussed this with marino already and atm the only solution it swiping to hide, but then the user cannot bring back the banner, **which is why maybe an icon next to the plus button (on Files app) can bring it back**

- Currently the banner is at the bottom, which is better than at the top, but it will still hide the last files (or other interactables) if scrolling to the bottommost part of the scroll view
image

- IMO "tap to cancel and swipe to hide" is way too complicated for regular users, as it kinda sounds the same. Just tap to hide is enough. However @marinofaggiana mentioned cancelling in Transfers is not enough as some steps begin outside of Transfers, so not sure about that.

cc @Ivansss @SystemKeeper

PR #3895, opened 2025-11-27
Fix empty files not being able to be shared

mpivchev:
An empty .md or .txt is 0 bytes, and the check if the file exists in the document storage expects the size of the file to be > 0. This removes the check, which allows a 0 byte file to be shared. 

@marinofaggiana do you think this is a dangerous solution? I haven't noticed any issues so far. I think all the other checks we do to make sure the file exists are enough.

### DIFF
--- a/iOSClient/Utility/NCUtilityFileSystem.swift
+++ b/iOSClient/Utility/NCUtilityFileSystem.swift
@@ -178,7 +178,7 @@ final class NCUtilityFileSystem: NSObject, @unchecked Sendable {
             return (fileNameViewSize == metadata.size) && metadata.size > 0
 #else
             if metadata.isDirectoryE2EE == true {
-                if (fileNameSize == metadata.size || fileNameViewSize == metadata.size) {
+                if fileNameSize == metadata.size || fileNameViewSize == metadata.size {
                     return true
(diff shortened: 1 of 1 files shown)


marinofaggiana:
#if EXTENSION_FILE_PROVIDER_EXTENSION
            return (fileNameViewSize == metadata.size) && metadata.size > 0
#else 

Must exists because  in FILE_PROVIDER_EXTENSION I'm not sure of the effects

mpivchev:
> #if EXTENSION_FILE_PROVIDER_EXTENSION return (fileNameViewSize == metadata.size) && metadata.size > 0 #else
> 
> Must exists because in FILE_PROVIDER_EXTENSION I'm not sure of the effects


Done, please recheck

mpivchev:
> > #if EXTENSION_FILE_PROVIDER_EXTENSION return (fileNameViewSize == metadata.size) && metadata.size > 0 #else
> > Must exists because in FILE_PROVIDER_EXTENSION I'm not sure of the effects
> 
> Done, please recheck

@marinofaggiana

mpivchev:
This issue seems to be gone so this PR is not needed.

Issue #3932, opened 2025-12-17
iOS - Login with deep links or Passkeys do not work when used in OAuth 2.0 flow


UBaggeler:
### Steps to reproduce
1. Server: Configure OAuth 2.0 client under "Administration" > "Security" with provider that uses deep links (custom url scheme or universal links) or Passkeys in their login flow
2. iOS App: Start login, go to OAuth 2.0 provider
3. Try to tap on link with deep link (e.g. "superloginapp://") or use Passkeys

### Expected behaviour
App should allow to do deep links or use Passkeys

### Actual behaviour
Login view (WKWebview) does not handle deep links or Passkeys. These are restrictions of WKWebViews (which are not recommended for login flows)

### Environment data

**iOS version:** any (tested on iOS 26.2)

**Nextcloud iOS app version:** 7.2.2 / master-branch

**Nextcloud version:** independent

UBaggeler:
From looking at the history it seems that around 6.4.0 (https://github.com/nextcloud/ios/pull/3303) the in-app browser (using SFSafariViewController) and external system browser was removed and replaced in favor of an internal WKWebView. Looking at the PR and the issues, it is not clear why this change was introduced. My guess is (from interpreting release notes) that there were issues cookies, that did not get cleared (as SFSafariViewController and the system browser do not allow to clear cookies). 
Also it seems that Android is (still) opening the system browser.

Using WKWebView comes with limitations (handling of deep-links, Passkeys among other things) and it is generally advisable to not use (embedded) WKWebViews for logins on iOS (also from security perspective). 
With Apple's introduction of ASWebAuthenticationSession (backed by SFSafariViewController) it is also possible to start ephemeral sessions with an empty cookie store.

@mpivchev @marinofaggiana What was the reason to revert back to WKWebViews after introducing in-app browser and system browser login options before?

mpivchev:
https://github.com/nextcloud/ios/pull/3996

Issue #3891, opened 2025-11-26
iPadOS app won’t log in fully


kryton-me:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. install “Nextcloud” app
2. tap “log in”
3. enter URL
4. get screen saying “Please log in before granting Mozilla/5.0 Nextcloud-iOS 7.2.2 acess…
5. tap “Log in”
6. enter email & password, tap log in
7. enter authentication code, tap submit
8. enter email & password, tap log in (yes again)
9. enter authentication code, tap submit
10. screen says “Your browser is not supported…”, tap “Continue with this unsupported browser”
11. Get the web interface of nextcloud and no files integration

### Expected behaviour
Tell us what should happen

I assume I should get the same interface as the iOS app which I have working in iOS 18.7.2

### Actual behaviour
Tell us what happens instead, if possible also add a screenshot

as discribed in stepts to repeat
I don’t have firefox (Mozilla) installed on the iPad

### Logs

```
If applicable, you can post the iOS app or server logs (removing any sensitive information).

iOS logs with sensitive bits hopefully removed and not the error codes

2025-11-25 13:53:14 [ERROR] Login poll result for token "" is not successful!
2025-11-25 13:53:15 [INFO] Network response request: https://.nextcloud-ionos.com/index.php/login/v2/poll?token=, result: failure(Alamofire.AFError.sessionTaskFailed(error: Error Domain=NSURLErrorDomain Code=-1005 "The network connection was lost." UserInfo={_kCFStreamErrorCodeKey=53, NSUnderlyingError=0x135cefd50 {Error Domain=kCFErrorDomainCFNetwork Code=-1005 "(null)" UserInfo={_NSURLErrorNWPathKey=satisfied (Path is satisfied), viable, interface: en0[802.11], ipv4, dns, uses wifi, LQM: good, _kCFStreamErrorCodeKey=53, _kCFStreamErrorDomainKey=1}}, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <>.<331>, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataTask <>.<331>"
[...]

Issue #3804, opened 2025-10-02
iOS 26 “Your browser is not supported”


leotrost:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Update to iOS 26
2. Attempt login

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead, if possible also add a screenshot
Warning saying the browser is not supported no matter the default browser on iOS. Logs in to the website but does not redirect back to the app to allow for backups, etc

![Image](https://github.com/user-attachments/assets/c118f784-c00a-43b3-9735-89b0586a1e77)

### Logs

```
If applicable, you can post the iOS app or server logs (removing any sensitive information).
```

### Environment data

**iOS version:** e.g. iOS 26.0

**Nextcloud iOS app version:** 7.1.7

**Server operating system:** TrueNAS Scale (App store) 

**Web server:** nginx

**Database:**

**PHP version:**

**Nextcloud version:** 32.0.0

richevanscybermyte:
We are having the same issue. 

Nextcloud iOS app version: 7.1.7

Server operating system: Kubernetes

Web server: Apache

Database: Postgres

PHP version: PHP 8.3.24

Nextcloud version: docker.io/nextcloud:31.0.7-apache (official)

chenster101:
Having the same issue. Still can't use the app.

vauvenal5:
I have the same issue, it happens for me after the selection of which 2FA method I want to use. The 2FA is not triggered at all it just directly redirects back to the login mask. After entering the login credentials a 2nd time, the 2FA is accepted and then we end up with the unsupported browser dialog.

Workaround that works for me is using app passwords.

mpivchev:
Should be fixed with: https://github.com/nextcloud/ios/pull/3996

PR #3938, opened 2025-12-26
Improve login to use ASWebAuthenticationSession with integrated browser

UBaggeler:
This PR improves log-in experience for NextCloud servers by migrating to ASWebAuthenticationSession.

- Using in-app system/default browser to restrictions/shortcomings of WKWebView
- Setting `prefersEphemeralWebBrowserSession` makes sure we always start fresh (without any cookies from previous sessions or default browser)
- Using callback to keep `nc://login` redirect handling intact (pre V2 flow)
- mTLS supported (can be configured in iOS Settings.app)

Fixes issues with deep-links and passkeys: https://github.com/nextcloud/ios/issues/3932
Fixes "Your browser is not supported": https://github.com/nextcloud/ios/issues/3891 https://github.com/nextcloud/ios/issues/3804

### DIFF
--- a/iOSClient/Login/NCLogin.swift
+++ b/iOSClient/Login/NCLogin.swift
@@ -8,7 +8,7 @@ import UIKit
 import NextcloudKit
 import SwiftEntryKit
 import SwiftUI
-import SafariServices
+import AuthenticationServices
 
 class NCLogin: UIViewController, UITextFieldDelegate, NCLoginQRCodeDelegate {
     @IBOutlet weak var imageBrand: UIImageView!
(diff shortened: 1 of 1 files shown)


mpivchev:
Hi @UBaggeler, thanks for the PR! We will review it soon.

mpivchev:
Hi @UBaggeler, thank you again for this! <3 

I will repurpose this PR into a new rebased one that includes your `ASWebAuthenticationSession` implementation, as it adds the great benefit for passkeys. I will of course credit your implementation and ideas proposed here.

However, as far as I'm aware (and tested on local), certificates from Settings are still not taken into account, as every app, and the internal browser we use, has their own sandbox, and it cannot read any certificates from iOS.

Because of this we need to handle mTLS certificates ourselves, using our own app sandbox, and is why I made this (in progress) PR that allows uploading and using mTLS certificates from the app itself: https://github.com/nextcloud/ios/pull/3685. Please let me know if you have any ideas and/or improvements on this!

Regarding the v1 login flow, it is officially deprecated, so that part of the PR sadly cannot be merged.

mpivchev:
Hi, here is the PR: https://github.com/nextcloud/ios/pull/3996

PR #3978, opened 2026-02-09
Assistant Chat

mpivchev:
Prerequisite:  https://github.com/nextcloud/NextcloudKit/pull/209
Implements https://github.com/nextcloud/files-clients/issues/111

image


marinofaggiana:
- Do not update (pool) after a query to assistant ?

PR #4025, opened 2026-03-11
New app icon

mpivchev:
Replaced multi icon asset with a single icon that is auto scaled by iOS.
Used a new liquid glass icon created via Icon Composer and exported as a .png.


mpivchev:
@Ivansss More context:

Replace multi icon with single icon in asset catalog.

image

We cannot use an Icon Composer .icon file for now because of brander, but at some point I assume asset catalog icon will be deprecated.

marinofaggiana (approved):

PR #4032, opened 2026-03-13
Media queue and Picture-in-Picture support

s-kaliberda:
Hi team,

After discussing media queue and Picture-in-Picture in the [pull request](https://github.com/nextcloud/ios/pull/4007), we agreed to use two libraries for media playback so that Picture-in-Picture will work. The media player uses AVKit for videos when the format is supported; otherwise, it uses MobileVLCKit.

This approach allowed us to preserve all existing functionality while also adding new features, such as a media queue for audio files and Picture-in-Picture for video files.

We implemented this logic in this pull request, and we’d like to hear your opinion on the implementation. Could you review the code and share your feedback?

### DIFF
--- a/Nextcloud.xcodeproj/project.pbxproj
+++ b/Nextcloud.xcodeproj/project.pbxproj
@@ -11,8 +11,12 @@
 		2C1D5D7923E2DE9100334ABB /* NCBrand.swift in Sources */ = {isa = PBXBuildFile; fileRef = F76B3CCD1EAE01BD00921AC9 /* NCBrand.swift */; };
 		2C33C48223E2C475005F963B /* NotificationService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C33C48123E2C475005F963B /* NotificationService.swift */; };
 		2C33C48623E2C475005F963B /* Notification Service Extension.appex in Embed Foundation Extensions */ = {isa = PBXBuildFile; fileRef = 2C33C47F23E2C475005F963B /* Notification Service Extension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
-		2F96A1BAFB10ACFEAC68EF1C /* NCContextMenuPlayerTracks.swift in Sources */ = {isa = PBXBuildFile; fileRef = B4C7A5B36D1ED178FB6B76CB /* NCContextMenuPlayerTracks.swift */; };
 		370D26AF248A3D7A00121797 /* NCCellMain.swift in Sources */ = {isa = PBXBuildFile; fileRef = 370D26AE248A3D7A00121797 /* NCCellMain.swift */; };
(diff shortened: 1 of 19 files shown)


mpivchev:
Hi @s-kaliberda, we will update you soon.

mpivchev:
Hi @s-kaliberda, would you mind writing an overview of how this implementation works, as well as what would be good to test.
Also, if possible, it would be nice to document parts of the code wherever it matters.

mpivchev:
What do I have to do to enable PiP? I don't see an option for this .mov file.

image

s-kaliberda:
Hi. PiP doesn't work on iPhone simulator, use physical device or iPad simulator.

s-kaliberda:
@mpivchev I made a new commit with fixes to some bugs we found during our testing, please, pull the latest changes.

mpivchev:
Thanks!

Some things I noticed:

1. When I have PIP on and the app is minimized, if I press the maximize button on the PIP, it pauses the video. The video should continue playing.
2. If you turn on PIP and you go to another folder in the app, then click the maximize button on the PIP, the player screen opens, but the app does not switch to the folder that the playing file is in. Is this intended? Not sure if it's better to auto switch back to the playing file folder (may be bad for UX) or never switch. @marinofaggiana what do you think?
3. Switching to PIP still shows the video in the main frame with the PIP icon and "This video is showing in PIP" descrip'tion on top. The main frame should be black and only show the icon and description:
IMG_1606

I will continue testing.

marinofaggiana:
First test - when go back the video continue in background - I strongly recommend waiting for our refactoring of Viewer.

mpivchev:
Hi 👋 As @marinofaggiana mentioned earlier, we have a rewrite of the viewer component in the app: https://github.com/nextcloud/ios/pull/4118

Please review it and check if all features you need are there, and if not, you can rebase and implement the additional features from this PR. Please let us know if you need any help :)

marinofaggiana:
@s-kaliberda We've done a lot of refactoring, you can test the version in Master 33.1.0 thanks

marinofaggiana:
close it why already completed. thanks.

s-kaliberda:
Hi, I'm sorry that I'm responding late.
Are you going to implement media queue?

marinofaggiana:
I have rewrite the the viewer and a lot of Media alg. with several news.

s-kaliberda:
I agree with you; it can be time-consuming to implement a media queue.
So, for now, Nextcloud doesn't have a media queue, and background playback works only when the media viewer is opened inside the application.
Media queue and background playback were present in this pull request.
Thank you for adding Picture-in-Picture.

marinofaggiana:
media queue ? what do you mean

s-kaliberda:
We implemented a media queue for audio files. When we open an audio file from a folder with a few audio files, after finishing playing one audio file, the player moves to another.

marinofaggiana:
Now it's a very easy logic to reproduce but I'm not convinced that everyone expects it...

PR #4058, opened 2026-04-08
Add in-app camera with direct upload to Nextcloud

Rasmussw:
This PR adds a custom in-app camera to the iOS client that captures photos and videos and uploads them directly to Nextcloud, bypassing the device camera roll by default. Users can optionally enable saving to the camera roll via a per-session toggle, with a global default available in Settings -> Advanced.

In the current flow, media is always saved to the system photo library before being uploaded to Nextcloud. This creates potentially unwanted local copies, introduces privacy concerns, and adds manual cleanup for users. This change ensures media can be captured and stored directly in Nextcloud without ever touching the camera roll. This is especially important in regulated environments with BYOD-policies, i.e. environments where media content captured can identify a person, thus potentially violating GDPR.

The custom camera uses AVFoundation to capture media directly to a temporary directory. Files are reviewed before upload and then passed into Nextcloud's existing upload pipeline. Saving to the camera roll is fully optional and only occurs if explicitly enabled by the user. Existing picker and upload flows remain unchanged.

Additional notes
Default behaviour does not affect existing users
Upload reliability uses existing background upload infrastructure
Camera roll is never accessed unless explicitly enabled


marinofaggiana:
@Rasmussw can you make a rebase ? thanks

marinofaggiana (changes requested):
Please make a rebase

Rasmussw:
Hi @marinofaggiana I've rebased onto the latest master.

marinofaggiana:
> Hi @marinofaggiana I've rebased onto the latest master.

There are some compilation errors (due to the master change). Please rebase and try compiling it. Thanks.

marinofaggiana:
Screenshot 2026-04-28 alle 09 26 17

marinofaggiana (changes requested):
do not build yet

iOSClient/Settings/NCPreferences.swift
```diff
@@ -203,6 +204,15 @@ final class NCPreferences: NSObject {
         }
     }
 
+    var saveCameraMediaToCameraRoll: Bool {
+        get {
+            return getBoolPreference(key: "saveCameraMediaToCameraRoll", defaultValue: false)
```
marinofaggiana: default, for compatibility with previous version, must be set to true

iOSClient/Supporting Files/en.lproj/Localizable.strings
```diff
@@ -474,6 +479,7 @@
 "_e2e_remove_folder_encrypted_"     = "Decrypt";
 "_e2e_file_encrypted_"              = "File encrypted";
 "_e2e_goto_settings_for_enable_"    = "This is an encrypted directory, go to \"Settings\" and enable end-to-end encryption";
+"_e2e_error_"                       = "An internal end-to-end encryption error occurred.";
```
marinofaggiana: ???

iOSClient/Main/Create/Upload Assets/NCUploadAssetsModel.swift
```diff
@@ -1,55 +1,121 @@
...
 
+// MARK: - CameraAssets helper
+enum CameraAssets {
+    struct TempAsset {
+        let fileURL: URL
+        let fileName: String
+        let isVideo: Bool
+    }
+}
+
```
marinofaggiana: it's used ?

marinofaggiana (changes requested):
read first comments

marinofaggiana:
iPad before:

IMG_0016

iPad after:

IMG_0017


**In addition to the incorrect view incompatible with iPad, on the second attempt it did not display the image (black)**

marinofaggiana:
Please use Lint, the code should be well formatted and commented in any modified places.

Screenshot 2026-05-05 alle 07 50 02

marinofaggiana:
_Note_: dear @Rasmussw  It seems like the change is significant. 
Why wasn't a fork on TLPhotoPicker considered to make the necessary changes there so as not to have to heavily intervene on the Nextcloud side of the process?

The library is very outdated and one might consider replacing it (also).

Rasmussw:
Hi Marino Thanks for the review.
First of all, we need to understand if there is an actual interest in this
feature from Nextcloud, if we take care of the issues you mentioned? If
there is, is it then a requirement that we fork TLPhotoPicker or is fine
the way we have implemented the feature as it is now? We believed that
there would be more work and maintenance for you at Nextcloud if we
implemented the feature by making the fork, thus we went with a solution
matching the current state of the Nextcloud iOS app.
We will off course look into the problems regarding the lint and ipad
issues.

Best regards, Rasmus

marinofaggiana:
Hi @Rasmussw ,

Thanks for the clarification.

Yes, I am interested in this feature, and I think the proposed change is a very good improvement for the app.

From my point of view, keeping most of the implementation on the Nextcloud side is fine and probably easier to maintain. My main concern was the automatic save to the user’s Photo Library when taking a photo or video from TLPhotoPicker.

Since we have now identified the exact point where this happens, I think the cleanest solution would probably be a small and focused change in TLPhotoPicker: add a configuration flag to decide whether captured photos/videos should be saved to the Photo Library or written to a temporary/custom location instead.

On the Nextcloud side, we would then only need to retrieve those captured images/videos from that location and continue with the existing upload flow.

However, I also noticed that the camera capture UI/flow seems to have changed, especially on iPad. I would like to better understand that part before deciding the final direction, because this is not only about where the captured file is saved, but also about how the capture view is presented and handled across devices.

So, I would not require a full fork if this can be handled with a small, focused change. Ideally, this could be proposed upstream. If that is not possible, then a minimal fork limited to this behavior could also be acceptable.

We will of course also need to address the lint and iPad issues before merging.

Rasmussw:
Hi Marino
Thanks for the response.
We will look into what is possible and get back to you.

marinofaggiana:
Definitely better, but I don't understand. As a dependency, you use https://github.com/tilltue/TLPhotoPicker 2.0.0 < 3.0.0, so the feature is already present?

Doing a quick test, I saw that the file name is UUID, which breaks the normal file naming convention. It would be a good idea to use the same mask.

( let destURL = tempDir.appendingPathComponent(UUID().uuidString + ".jpg")

Rasmussw:
Hi @marinofaggiana ,

Thanks for the review.

To answer your question, didCaptureMediaURL does not exist in the original TLPhotoPicker. I added it in my own fork and submitted a PR upstream (https://github.com/tilltue/TLPhotoPicker/pull/383), which is still awaiting review. The current committed code actually still points to the original tilltue/TLPhotoPicker.

I will fix the UUID filename issue, but regarding the dependency, what would you prefer?

1. Point to my fork until the upstream PR is merged, then switch back to the original.
2. Wait for the upstream PR to be merged before moving forward with this PR.

Best regards,
Rasmus

marinofaggiana:
> Hi @marinofaggiana ,
> 
> Thanks for the review.
> 
> To answer your question, didCaptureMediaURL does not exist in the original TLPhotoPicker. I added it in my own fork and submitted a PR upstream ([tilltue/TLPhotoPicker#383](https://github.com/tilltue/TLPhotoPicker/pull/383)), which is still awaiting review. The current committed code actually still points to the original tilltue/TLPhotoPicker.
> 
> I will fix the UUID filename issue, but regarding the dependency, what would you prefer?
> 
> 1. Point to my fork until the upstream PR is merged, then switch back to the original.
> 2. Wait for the upstream PR to be merged before moving forward with this PR.
> 
> Best regards, Rasmus

Hi @Rasmussw,
did you ping it to see if it will be added to master?

Rasmussw:
Hi @marinofaggiana,

Good news, the upstream PR has actually just been merged into master(https://github.com/tilltue/TLPhotoPicker/pull/383), and a new tag(2.1.20) has been released with the changes. I will update the dependency to point to the official repositorys new version and push the UUID filename fix in my next commit.

Best regards, Rasmus

Rasmussw:
Hi @marinofaggiana 
  
Updated the PR with the following changes:

- Now using didCaptureMediaURL from the official TLPhotoPicker 2.1.21
- Filenames follow Nextcloud's naming convention via createFileName
- Rebased onto latest master
- Lint is clean

Best regards, Rasmus

marinofaggiana:
@Rasmussw can you make a git merge origin/master ? Sorry but I've been very busy

Rasmussw:
Hi @marinofaggiana 

Latest master is merged into the branch.

marinofaggiana (approved):

marinofaggiana:
@Rasmussw great job ! thank you !

Rasmussw:
Hi @marinofaggiana 

Thanks again for the review and for merging the PR!

Do you have an idea of when this will be included in a release/tag and available in the App Store?

Best regards Rasmus

marinofaggiana:
It won't be anytime soon; we've just added a new image and video viewer, as well as an updated Media search, which deserves a thorough test. You can try it out now! It's already in Master and TestFlight.

nwp-projectmanager:
Hi @marinofaggiana 

We're happy this PR has been merged and released.

What are the chances that a PR proposing a change in the default behaviour for this feature will be accepted? I.e. that when a users takes a photo, photos will not be stored locally in the camera roll by default before the flip is switched in the settings menu.

I ask because we might have overlooked this potential important distinction, that this should be the opposite way around to impose more privacy by default, rather than an option to set privacy by default by yourself.

Kind regards

Niels William

marinofaggiana:
Hi, sorry, I don't understand what you mean.

nwp-projectmanager:
I sent you an email with an idea to resolve my not-so-easy-to-understand message.

PR #4096, opened 2026-05-05
Focused auto upload

mpivchev:
- Adds a focused auto upload section. Allows to keep screen black while app is in foreground for faster upload.
- Add animated cloud icon when auto upload is on
- Use "X items remaining" as a counter when auto upload is on
- Move auto upload section to NCMore
- Fix invalid Scheduled + In Progress counter
- Add gradient to quota in NCMore

IMG_1697

IMG_1698

IMG_1699

IMG_1700


iOSClient/Data/NCManageDatabase+AutoUpload.swift
```diff
@@ -106,6 +106,39 @@ extension NCManageDatabase {
...
+                        global.metadataStatusUploadingAllMode)
+
+            guard !excludedIds.isEmpty else {
+                return results.count
+            }
+
+            return results
+                .filter("NOT (ocIdTransfer IN %@)", Array(excludedIds))
+                .count
+        } ?? 0
```
mpivchev: We filter out auto upload success files. Is that ok? @marinofaggiana

marinofaggiana:
- title button, I would put:  Auto upload photos \n
                                           1.649 items left 

1569 - 1649 ? mmm maybe tre webdav directory creation ?

image: total black it has never been used...

the touch on button works only on title and the ">" is not correct, it should be a normal button not a push


- when activate the auto upload I would put the <Focused Auto Upload> on Top of view, or an button on center screen with the detail like the below "Turn off auto upload" 

- when the passcode is activated the "focused" it turns off, which might be fine but I don't know what happens to the routine, please check if it works with the passcode request and with the function "Hide content when app is inactive"


To do:

 - Remove Keep screen awake option, as it's not needed anymore.

I would leave it anyway, is it in conflict or what?

marinofaggiana (changes requested):
look note.

mpivchev:
1. "image: total black it has never been used..." what do you mean?
2. "the touch on button works only on title and the ">" is not correct, it should be a normal button not a push" Fixed.
3. "when activate the auto upload I would put the on Top of view, or an button on center screen with the detail like the below "Turn off auto upload"" Done
4. "when the passcode is activated the "focused" it turns off, which might be fine but I don't know what happens to the routine, please check if it works with the passcode request and with the function "Hide content when app is inactive" You are not asked for passkey when returning from the focus mode.

mpivchev:
"1569 - 1649 ? mmm maybe tre webdav directory creation ?"
Actually it makes no sense why folder creation and other similar operations are included in the count?
Transfers screen and number should only be for each file that is queued to be uploaded. Folder creation and so on do not matter?

iOSClient/Networking/NCNetworkingProcess.swift
```diff
@@ -126,12 +126,10 @@ actor NCNetworkingProcess {
         currentAccount = account
     }
 
-    private func inWaitingDownloadUploadCount() async -> Int {
-        let countTransferDownloadingUploadingSuccess = await NCNetworking.shared.metadataTranfersSuccess.count(statuses: NCGlobal.shared.metadatasStatusDownloadingUploading)
```
mpivchev, line 130: `await NCNetworking.shared.metadataTranfersSuccess.count(statuses: NCGlobal.shared.metadatasStatusDownloadingUploading)`
actually checks when the in progress metadatas are done and flushed out, it does not check the **actual** in progress metadatas.
After this fix, we now count it properly.

Case 1:
If transfers screen shows 10 Scheduled and 10 In Progress:
1. Badge shows 20
2. Auto upload counter shows 20

Case 2:
If transfers screen shows 10 Scheduled (5 auto upload metadatas) and 10 In Progress (5 auto upload metadatas):
1. Badge shows 20
2. Auto upload counter shows 10

@marinofaggiana

marinofaggiana: metadataTranfersSuccess are success but not yet write in DB (cache). 
Technically, to make a calculation we could also not include them and look ONLY in the DB

mpivchev: > metadataTranfersSuccess are success but not yet write in DB (cache). Technically, to make a calculation we could also not include them and look ONLY in the DB

That's what I'm doing already, unless you mean something else.

marinofaggiana (approved):

Nextcloud.xcodeproj/project.pbxproj
```diff
@@ -5015,7 +5043,7 @@
 		};
 		F7C55C872FB5AEF7004A974F /* PBXTargetDependency */ = {
 			isa = PBXTargetDependency;
-			target = F7C55C792FB5AEF7004A974F /* Action Assistant */;
+			target = F7C55C792FB5AEF7004A974F /* Assistant Action */;
```
mpivchev, line 5046: This naming matches all the other targets. Bundle ID is not changed @marinofaggiana

PR #4108, opened 2026-05-12
Feature/eurooffice direct editing

moodyjmz:
The direct editing code was comparing against editor display names (e.g. "Nextcloud Text") rather than their API identifiers (e.g. "text"). This worked by coincidence for onlyoffice since the name and identifier happened to match, but broke for EuroOffice where the display name is "Nextcloud Office" and the identifier is "eurooffice". This PR introduces NCDirectEditorAdapter as a central registry keyed on identifiers, so any editor registered there works end-to-end without touching dispatch code in multiple files.


marinofaggiana:
moved https://github.com/nextcloud/ios/pull/4109

PR #4111, opened 2026-05-15
Fix VoiceOver labels in share extension folder picker

### LINKED ISSUE #4099: Accessibility: BLOCKING - Folder selection when sharing a file from another app not accessible with the VoiceOver screen reader
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Share any file from another application, by using the native share dialog and choosing "Nextcloud".
2. Swipe right/left with one finger while the VoiceOver screen reader is on.
3. (Optional) To see something closer to how it is supposed to work properly, cancel the process and navigate to the "Files" tab of the app itself.

### Expected behaviour

VoiceOver should read the same information present on screen, as in step 3 (regular "Files" tab).

### Actual behaviour

In step 2 (sharing screen after choosing "Nextcloud"), VoiceOver does not read the name or details of any folder, ahtough they are visible on screen and can be selected, both with and without VoiceOver. The actual issue is that a VoiceOver user cannot know which folder they are focusing in each sswipe.

due to this critical accessibility issue, users relying on VoiceOver, and potentially also on other assistive technologies which use programmatically available data in order to convey information to their users, cannot upload files autonomously by sharing them with the app. They have to get support either from a third person or from screen recognition features of the screen reader, which are often sluggish and not especially reliable.
this issue is in nature related to #3069, which is also blocking for users of assistive technologies but in a different screen and not exactly in the same way.

### Environment data

**iOS version:** iOS 26.3.1

**Nextcloud iOS app version:** 33.0.7.5

**Server operating system:**

**Web server:** 

**Database:**

**PHP version:**

**Nextcloud version:**

web3blind:
## Summary
- add an explicit accessibility label to folder cells in the share extension target-folder picker
- include the visible folder name and metadata text that is already shown in the cell
- expose folder rows as actionable elements for VoiceOver users

## Issue
Fixes #4099

## Testing
- Not run: iOS/Xcode build is not available in this environment
- Static inspection: verified the share extension uses NCListCell for target folders and now calls setAccessibility(label:value:) for each cell

## Accessibility impact
This should let VoiceOver announce the folder name/details while sharing a file into Nextcloud from another app, instead of focusing unnamed selectable rows.

### DIFF
--- a/Share/NCShareExtension+DataSource.swift
+++ b/Share/NCShareExtension+DataSource.swift
@@ -104,6 +104,13 @@ extension NCShareExtension: UICollectionViewDataSource {
 
         cell.setTags(tags: Array(metadata.tags))
 
+        let accessibilityLabel = [metadata.fileNameView, cell.labelInfo.text, cell.labelSubinfo.text]
+            .compactMap { $0?.trimmingCharacters(in: .whitespacesAndNewlines) }
+            .filter { !$0.isEmpty }
+            .joined(separator: ", ")
+        cell.setAccessibility(label: accessibilityLabel, value: "")
+        cell.accessibilityTraits = .button
+
         cell.separator.isHidden = collectionView.numberOfItems(inSection: indexPath.section) == indexPath.row + 1
 
         return cell


mpivchev:
Hi @web3blind, thank you for the PR!
I will get back to you soon.

mpivchev (approved):
LGTM

PR #4116, opened 2026-05-22
fix(AppIcon): Use app icon consistent with Talk

kra-mo:
Made by @yotamguttman

This is the agreed-upon style.


mpivchev (approved):
Thanks!

PR #4171, opened 2026-07-08
Odd plus button UI

### LINKED ISSUE #3892: Odd plus button on iOS 26
This looks quite unexpected.

Image

The button should be tinted instead and the icon itself should have no background color to avoid the circle in a circle. Example from reminders:

Image

marinofaggiana:
Screenshot 2026-07-08 alle 08 50 14


marinofaggiana:
@kra-mo , @jancborchardt

jancborchardt (commented):
Thank you, very nice! :) Looks much better

PR #4174, opened 2026-07-08
Rerequest push notifications

### LINKED ISSUE #4145: Push notifications are sometimes not up to date
There is a server bug that is circumvented on Android and Desktop, but not on iOS.

1. Bob creates a deck card.
3. A notification is created on server with "You have created a deck card". `userId` is Bob. `author = userId` is true.
4. This notification is sent to Bob, as he is the author.
5. However, since this notification is only made once, with Bob as `userId`, it is never changed, and is also sent as a push notification to all other sharees, still appearing as "You have created a deck card". This is incorrect, as the sharees are not the author Bob.

[Android](https://github.com/nextcloud/android/blob/6a8c4384c349c15a6167c038c512f91e97ddb711/app/src/main/java/com/nextcloud/client/jobs/NotificationWork.kt#L114-L115) and [Desktop](https://github.com/nextcloud/desktop/blob/a80ec7c8800840b275379ab9634a3e5bb936928b/src/libsync/pushnotifications.cpp#L111) alleviate this by receiving the original notification, but then rerequesting the same notification, with their own `userId` for the logged in client user. This allows the notification to be constructed again, with the proper string. In this case, Alice, which `author != userId`. The notification will then be "Bob created a deck card".

Ultimately this bug should be resolved server side, but to keep feature parity with other clients, iOS should also rerequest an up to date notification.

mpivchev:
NCKit: https://github.com/nextcloud/NextcloudKit/pull/231
Fixes https://github.com/nextcloud/ios/issues/4145

Rerequest each notification from server so it's up to date. This resolves issue where a notification would show the wrong content.

- Tested with multiple accounts, 2 accounts from same instance and 1 from a different instance. Notifications behave as usual, but have updated content if needed.
- Tested NCKit timeout + `serviceExtensionTimeWillExpire` timeout

### DIFF
--- a/Notification Service Extension/NotificationService.swift
+++ b/Notification Service Extension/NotificationService.swift
@@ -58,11 +58,11 @@ class NotificationService: UNNotificationServiceExtension {
                     nkLog(debug: "🔑 Loaded private key for \(tableAccount.account): prefix(Base64)=\(prefixBase64)")
 
                     guard let decryptedMessage = NCPushNotificationEncryption.shared().decryptPushNotification(message, withDevicePrivateKey: privateKey) else {
-                        bestAttemptContent.body = "Error decryption for \(tableAccount.account)"
+                        bestAttemptContent.body = "Error decrypting notification for \(tableAccount.account)"
                         continue
                     }
                     guard let data = decryptedMessage.data(using: .utf8) else {
-                        bestAttemptContent.body = "Error decryption data utf8 for \(tableAccount.account)"
+                        bestAttemptContent.body = "Error decrypting UTF8 notification data for \(tableAccount.account)"
                         continue
                     }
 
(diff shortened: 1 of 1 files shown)


Notification Service Extension/NotificationService.swift
```diff
@@ -35,59 +37,101 @@ class NotificationService: UNNotificationServiceExtension {
...
-        // Use this as an opportunity to deliver your "best attempt" at modified content, otherwise the original push payload will be used.
-        if let contentHandler = contentHandler, let bestAttemptContent = bestAttemptContent {
-            bestAttemptContent.title = ""
-            bestAttemptContent.body = "Nextcloud Notification Time Will Expire"
-            contentHandler(bestAttemptContent)
+        deliver()
+    }
+
+    private func deliver() {
+        let handler = deliveryLock.withLock {
```
mpivchev, line 127: Lock is added as `serviceExtensionTimeWillExpire` comes from a different thread: https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension/serviceextensiontimewillexpire()

marinofaggiana (approved):

PR #4200, opened 2026-07-23
feat: Unified sharing

mpivchev:
- Implements: https://github.com/nextcloud/NextcloudKit/pull/238
- Share tabs now use SwiftUI segmented component
- Removed Parchment lib


marinofaggiana:
https://github.com/nextcloud/ios/pull/4291

PR #4271, opened 2026-08-20
refactor: reorganize plus menu document actions

marinofaggiana:
Screenshot 2026-08-20 alle 15 22 30

Screenshot 2026-08-20 alle 15 18 02

Screenshot 2026-08-20 alle 15 23 58

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI


kra-mo (changes requested):
Folder/Encrypted folder should be regular (vertical) entries instead of horizontal ones.

Otherwise, the strings provided by the app should match those of the mockup.

PR #4275, opened 2026-08-26
fix: open custom unified search file results

### LINKED ISSUE #4274: Custom unified search file results cannot be opened
### Steps to reproduce

1. Register a custom unified search provider on a Nextcloud server.
2. Return a `SearchResultEntry` that refers to a readable file in the user's Nextcloud storage and includes the standard file attributes:

   ```php
   $entry->addAttribute('fileId', (string)$file->getId());
   $entry->addAttribute('path', $userRelativePath);
   ```

3. Search for the entry in Nextcloud iOS.
4. Tap the result shown under the custom provider.

### Expected behaviour

The result is represented as a Nextcloud file and opens in the native viewer, in the same way as a result from the built-in `files` provider.

### Actual behaviour

The result is displayed with a generic URL/file icon, but tapping it has no visible effect. No follow-up network request is made and no error is shown or logged.

The same OCS search response works in the web client and opens natively in the Android client.

### Logs

The unified search request completes successfully. Tapping the result produces no subsequent request or log entry.

`NKSearchEntry` parses the standard `fileId` and `path` attributes for every provider. However, `NCCollectionViewCommon+Search.swift` only resolves them to real file metadata when the provider ID is exactly `files`. Entries from every other provider are converted to temporary URL metadata with a random `ocId`.

When the entry is selected, `NCCollectionViewCommon+CollectionViewDelegate.swift` calls `setMetadataSessionInWaitDownloadAsync` before opening the URL. The temporary metadata was never stored in Realm, so this method returns `nil` and the selection handler exits silently.

Custom providers that point to a file already stored in Nextcloud should be allowed to opt into native file handling through `fileId`/`path`, independently of their provider ID. URL-only results should retain the existing fallback behaviour.

### Environment data

**iOS version:** 26.6

**Nextcloud iOS app version:** 34.1.4

**Server operating system:** Official Nextcloud Docker image

**Web server:** Apache

**Database:** SQLite (isolated reproduction instance)

**PHP version:** 8.4.24

**Nextcloud version:** 33.0.8

Dennis-Otto:
## Summary

- Resolve custom unified search entries with `fileId` or `path` attributes to native Nextcloud file metadata.
- Keep the existing URL fallback for results that do not refer to a Nextcloud file.
- Preserve the existing behaviour of the built-in `files` and `fulltextsearch` providers.

## Reasoning

Nextcloud iOS currently creates temporary URL metadata for every result from a custom unified search provider, even when the entry exposes the standard `fileId` and `path` attributes. Selecting that temporary item exits silently because its random `ocId` is not present in Realm.

Resolving those attributes before creating the URL fallback makes file-backed results use the regular native file metadata and viewer. It also gives them their actual file type and preview instead of a generic URL icon.

Closes #4274.

## Testing

- Reproduced with Nextcloud iOS 34.1.4 against an isolated Nextcloud 33.0.8 instance and a minimal custom search provider.
- Confirmed that the provider response contains a readable file ID and user-relative path.
- Ran `git diff --check`.
- Parsed the changed Swift source with `xcrun swiftc -frontend -parse`.
- A full Xcode build was not available locally; the project CI will perform the supported simulator build and test suite.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/iOSClient/Main/Collection Common/NCCollectionViewCommon+Search.swift
+++ b/iOSClient/Main/Collection Common/NCCollectionViewCommon+Search.swift
@@ -283,6 +283,21 @@ extension NCCollectionViewCommon {
 
         default:
             for entry in searchResult.entries {
+                if let fileId = entry.fileId,
+                   let metadata = database.getMetadataFromFileId(String(fileId)) {
+                    metadata.section = provider.name
+                    metadatas.append(metadata)
+                    continue
+                }
+
+                if let filePath = entry.filePath,
+                   let metadata = await loadMetadata(session: session,
+                                                     provider: provider,
+                                                     filePath: filePath) {
+                    metadatas.append(metadata)
+                    continue
+                }
+
                 let metadata = await NCManageDatabaseCreateMetadata().createMetadataAsync(
                     fileName: entry.title,
                     ocId: NSUUID().uuidString,


marinofaggiana:
thanks for idea: https://github.com/nextcloud/ios/pull/4325

PR #4288, opened 2026-09-09
fix(auto-upload): reduce JPEG compression quality to 0.85

### LINKED ISSUE #4287: Auto Upload: JPEG conversion uses compressionQuality 1.0, tripling file size (regression in #4223)
### Steps to reproduce
1. iOS Camera set to "High Efficiency" (HEIC), 24 MP (never manually changed in iOS)
2. Enable Auto Upload with compatibility mode (HEIC -> JPEG)
3. Compare the uploaded JPEG size before and after app version 34.1.2

### Actual behaviour
Same source photo (5712x4284, ~7 MB HEIC) now produces a ~23 MB JPEG.
Before 34.1.2 the same kind of photo produced ~8 MB. (first observed at around 2026-08-26)

### Expected behaviour
Converted JPEGs should stay in a reasonable size range. Quality 0.8-0.9 is
visually indistinguishable here and produces roughly a third of the filesize.

### Cause
In `iOSClient/Utility/NCCameraRoll.swift`, commit 82634b52d6 (PR #4223)
added `options: [compressionQuality: 1.0]` to the `jpegRepresentation` call.
Previously no options were passed and Core Image's default was used.

Quality 1.0 forces the encoder into a very inefficient mode, and since the
source is an already-compressed HEIC, its compression artifacts get encoded
into the JPEG as well. Apple confirms this behaviour and recommends ~0.8:
https://developer.apple.com/forums/thread/793946

### Suggestion
Lower the constant to 0.85, or expose it as a setting next to the existing
compatibility-mode toggle.

### Environment
- iPhone 16, iOS 26.6.1
- Nextcloud iOS app: 34.1.4.1
- Nextcloud server: 34.0.3

Hope this helps :-)

naelfe:
Quality 1.0 forces an inefficient encode and re-encodes the source HEIC's compression artifacts, roughly tripling the resulting file size with no visible gain.

Fixes #4287

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### Testing

Not built locally — no macOS machine available. Relying on CI.

The value 0.85 is a conservative pick; Apple's guidance suggests ~0.8 for
re-encodes of already-compressed sources:
https://developer.apple.com/forums/thread/793946

Scope: this is meant as a minimal fix so the regression stops shipping.
Happy to adjust the value if you'd prefer a different one. A user-facing
setting next to the compatibility-mode toggle would be a good follow-up,
but I'd rather leave that to a separate PR than expand this one.

### Note on the AI Usage

The one-line edit was made by hand. AI assistance was used for research —
locating the regression via `git blame` — and for drafting the issue and
commit text.

### DIFF
--- a/iOSClient/Utility/NCCameraRoll.swift
+++ b/iOSClient/Utility/NCCameraRoll.swift
@@ -345,7 +345,7 @@ final class NCCameraRoll: CameraRollExtractor {
                   let jpegData = CIContext().jpegRepresentation(
                     of: ciImage,
                     colorSpace: colorSpace,
-                    options: [compressionQuality: 1.0]
+                    options: [compressionQuality: 0.85]
                   )
             else {
                 throw NSError(domain: "ExtractAssetError", code: 3, userInfo: [NSLocalizedDescriptionKey: "JPEG conversion failed"])


Croydon:
I don't understand this. Nextcloud should upload files as-is and not change them at all. Am I missing something?

Edit: Alright, this is just when enabling this "compatible mode", I guess.

naelfe:
> I don't understand this. Nextcloud should upload files as-is and not change them at all. Am I missing something?
> 
> Edit: Alright, this is just when enabling this "compatible mode", I guess.

If I were to upload images as heic 1:1, then yes (I suppose).
But I use the compatibility option to upload as jpg, not heic.

mpivchev:
@marinofaggiana future JPEG conversion changes must be notified here for the server HEIC -> JPEG feature request https://github.com/nextcloud/server/issues/63754

PR #4292, opened 2026-09-11
Remove photo from Albums

harshada-15-tsys:
fix - Added the handling for removing photo from album
This will only remove photo from the album not delete.


marinofaggiana:
https://github.com/nextcloud/ios/pull/4293

Issue #4299, opened 2026-09-16
Background uploading / location updates issues


cmigliorini:
Hello @marinofaggiana @mpivchev,

Hi, I'm relying on iOS app a lot to upload pictures from my camera roll to my private Nextcloud instance, and have found it failing quite a lot, albeit intermittently. I also found I'm not the only one (https://help.nextcloud.com/t/nextcloud-ios-app-transfer-list-drains-without-sending-anything/246942/3). So I'm currently working on fixing this

- queue draining
- background transfers not triggered
- pictures not deleted from camera roll

Since I could not reliably describe or reproduce symptoms, and because although I'm a seasoned developer/CTO, I don't know much about iOS/Swift, I took upon myself to use Claude for help. I have been thouroughly experimenting (so far I think I walked about 10-20km just for the sake of testing this) and I think I'm closing in.

Would you be ok if I submitted a Claude-authored, reviewed as much as I can by myself, but most of all carefully tested, PR?

cmigliorini:
I just found out this https://github.com/nextcloud/ios/issues/3838. Did you actually work on the auto-upload _mechanism_ per se? I only focused on that. Look, attached is what I came up with, there's not a lot of code modifications -- and a lot of verbose comments in [my branch](https://github.com/cmigliorini/nextcloud-ios/tree/fix-upload-deletion-failures). What I can say is that now Nextcloud auto-upload seems very stable, I'm yet to test low power mode, though, and probably a few other things as well

[Nextcloud-iOS-ISSUE_DRAFT.md](https://github.com/user-attachments/files/32303973/Nextcloud-iOS-ISSUE_DRAFT.md)

marinofaggiana:
https://github.com/nextcloud/ios/issues/3838

PR #4301, opened 2026-09-16
Fix upload deletion failures

cmigliorini:
Submitting this as draft, see #4299. Still need to review some of it, and of course comply with all requirements, including sign-off, if this seems like something you want to accept.

## 🤖 AI (if applicable)

- [X ] The content of this PR was partly or fully generated using AI (Claude Opus)

### DIFF
--- a/iOSClient/Data/NCManageDatabase+Metadata.swift
+++ b/iOSClient/Data/NCManageDatabase+Metadata.swift
@@ -1303,7 +1303,13 @@ extension NCManageDatabase {
 
     func getAssetLocalIdentifiersUploadedAsync() async -> [String]? {
         return await core.performRealmReadAsync { realm in
-            let results = realm.objects(tableMetadata.self).filter("assetLocalIdentifier != ''")
+            // Only consider assets whose upload actually completed. Relying on the
+            // caller's transfer queue being globally empty as a proxy for "done" is
+            // unreliable (any unrelated pending download/upload starves cleanup) and,
+            // if that proxy were ever removed, this filter is what stands between a
+            // still in-flight upload and a premature deletion from the camera roll.
+            let results = realm.objects(tableMetadata.self)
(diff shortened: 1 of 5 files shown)


marinofaggiana:
Thanks for the work @cmigliorini 

Some of the fragile areas mentioned here are already well known to us, and we are currently testing the migration to the new PhotoKit background upload architecture.

Using AI to analyse the repository is not, by itself, the difficult part — we all have access to the same tools and can ask them to inspect the codebase and suggest changes.

The real engineering work is being able to critically evaluate those suggestions, understand their implications within the existing architecture, and take responsibility for the proposed solution.

That is particularly important in areas such as background execution, PhotoKit, lifecycle, concurrency and persistence, where code can look perfectly reasonable and still be architecturally wrong or introduce subtle regressions.

Before proposing changes of this scope, I would strongly recommend first gaining a deeper understanding of the relevant iOS APIs and of how this codebase is designed.

Otherwise, the difficult part is simply transferred to the maintainers, who then have to determine whether the AI-generated analysis and implementation are actually correct.

cmigliorini:
Hi @marinofaggiana , thanks for the reply. What my work amounted to is the testing itself, and it showed a few major flaws in the legacy mechanism, including -- that's pretty usual with this kind of things -- a few race conditions, and the fixes demonstrate improvements.

Now, if you're not considering maintaining the legacy autoupload, it makes sense to ignore this... I own iOS27 capable devices so if the upcoming version fixes these, then I'm not impacted and will be happy.

At any rate, I'm willing to participate in the beta testing, if you could include me to the group?

marinofaggiana:
Definitely! In fact, thank you.

cmigliorini:
Regarding APIs understanding, I have a few insights I'm happy to share, would you like me to do so here, in your PR, in your issue ?

PR #4318, opened 2026-09-19
fix: Only hide carousel for compact devices when in landscape

claucambra:
iPhone Duo is big enough to display the carousel

After:

Screenshot 2026-09-19 at 22 20 01

Before:

Screenshot 2026-09-19 at 22 23 56

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI (N/A)


mpivchev (approved):
Thanks!

